PR #21, opened 2022-03-30
feat: update library name

qingzhuozhen:
### Summary

- update library name according to discussion with @justin-fiedler 

### Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,7 +8,7 @@ ext {
     PUBLISH_DESCRIPTION = 'Amplitude Kotlin client-side SDK for Android'
     PUBLISH_GROUP_ID = 'com.amplitude'
     PUBLISH_VERSION = '0.1.0'
-    PUBLISH_ARTIFACT_ID = 'amplitude-analytics-android-sdk'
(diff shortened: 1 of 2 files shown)


android/build.gradle
```diff
@@ -8,7 +8,7 @@ ext {
     PUBLISH_DESCRIPTION = 'Amplitude Kotlin client-side SDK for Android'
     PUBLISH_GROUP_ID = 'com.amplitude'
     PUBLISH_VERSION = '0.1.0'
-    PUBLISH_ARTIFACT_ID = 'amplitude-android'
+    PUBLISH_ARTIFACT_ID = 'amplitude-analytics-android-sdk'
```
justin-fiedler, line 11: How about `analytics-android`? I agree with @bgiori the `amplitude` and `sdk` seem unnecessary.

justin-fiedler (changes requested):

android/build.gradle
qingzhuozhen: ok, let me update that for now. In that case, I will update SDK library to analytics-android as well

justin-fiedler (approved):

PR #25, opened 2022-04-10
feat: add event id support

qingzhuozhen:
### Summary

- add event id support

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
+++ b/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
@@ -58,14 +58,14 @@ class AndroidContextPlugin : Plugin {
         event.timestamp ?: let {
             val eventTime = System.currentTimeMillis()
             event.timestamp = eventTime
-            (amplitude as com.amplitude.android.Amplitude).lastEventTime = eventTime
+            getAndroidAmplitude().lastEventTime = eventTime
         }
         event.timestamp ?. let {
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
```diff
@@ -52,7 +56,18 @@ class AndroidContextPlugin : Plugin {
     private fun applyContextData(event: BaseEvent) {
         val configuration = amplitude.configuration as Configuration
         event.timestamp ?: let {
-            event.timestamp = System.currentTimeMillis()
+            val eventTime = System.currentTimeMillis()
+            event.timestamp = eventTime
+            (amplitude as com.amplitude.android.Amplitude).lastEventTime = eventTime
```
justin-fiedler, line 61: Maybe we can make a getter to avoid frequent type casting?

```
class AndroidPluginBase: Plugin {
  getAmplitude(): com.amplitude.android.Amplitude {
   return amplitude as com.amplitude.android.Amplitude
  }
}
```

kevinpagtakhan (approved):

android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
qingzhuozhen: I see, basically have the typecasting inside the plugin. That should be good for our own plugins and I can make that change. I wonder if in the future customer would need to do that typecasting in their plugins, and if we should switch to extension before that, or if they can do similar getter just for android.

PR #26, opened 2022-04-19
feat: remove json object from public interface

qingzhuozhen:
### Summary

- remove JSONObject from public interface

### Checklist

* Does your PR have a breaking change?:  yes

### DIFF
--- a/core/src/main/java/com/amplitude/core/Amplitude.kt
+++ b/core/src/main/java/com/amplitude/core/Amplitude.kt
@@ -101,10 +101,10 @@ open class Amplitude internal constructor(
      * @return the Amplitude instance
      */
     @JvmOverloads
-    fun track(eventType: String, eventProperties: MutableMap<String, Any?>? = null, options: EventOptions? = null): Amplitude {
+    fun track(eventType: String, eventProperties: Map<String, Any?>? = null, options: EventOptions? = null): Amplitude {
         val event = BaseEvent()
         event.eventType = eventType
-        event.eventProperties = eventProperties
+        event.eventProperties = eventProperties?.toMutableMap()
         options ?. let {
             event.mergeEventOptions(it)
         }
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -102,7 +101,7 @@ open class Amplitude internal constructor(
      * @return the Amplitude instance
      */
     @JvmOverloads
-    fun track(eventType: String, eventProperties: JSONObject? = null, options: EventOptions? = null): Amplitude {
+    fun track(eventType: String, eventProperties: MutableMap<String, Any?>? = null, options: EventOptions? = null): Amplitude {
```
bgiori, line 104: Does this need to be a `MutableMap`? Can we just use `Map`?

core/src/main/java/com/amplitude/core/utilities/JSON.kt
```diff
@@ -45,6 +45,17 @@ internal fun List<*>?.toJSONArray(): JSONArray? {
     return jsonArray
 }
 
+internal fun Array<*>?.toJSONArray(): JSONArray? {
```
bgiori, line 48: Nice. This makes me think we should change the `List<*>?.toJSONArray()` to `Collection<*>?.toJSONArray()` to cover more cases?

bgiori (changes requested):

core/src/main/java/com/amplitude/core/Amplitude.kt
qingzhuozhen: Only make this MutableMap as I adjust the eventProperties in BaseEvent as that. But I guess Map makes more sense for this interface when giving another thought. I will update this to Map.

core/src/main/java/com/amplitude/core/utilities/JSON.kt
qingzhuozhen: That's a good catch! Let me just adjust List to Collection then

justin-fiedler (approved):
I agree with @bgiori comments but otherwise LGTM.

core/src/main/java/com/amplitude/core/Amplitude.kt
bgiori: You should be able to use the `.toMutableMap()` extension function.

bgiori (approved):

PR #27, opened 2022-04-23
feat: add unit tests in id module

qingzhuozhen:
### Summary

- add unit tests in id module
- minor fixes

### Checklist

### DIFF
--- a/id/src/test/kotlin/com/amplitude/id/FileIdentityStorageTest.kt
+++ b/id/src/test/kotlin/com/amplitude/id/FileIdentityStorageTest.kt
@@ -1,0 +1,58 @@
+package com.amplitude.id
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Order
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class FileIdentityStorageTest {
+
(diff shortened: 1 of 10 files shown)


id/src/test/kotlin/com/amplitude/id/FileIdentityStorageTest.kt
```diff
@@ -0,0 +1,54 @@
...
+        val fileIdentityStorage = identityStorageProvider.getIdentityStorage(identityConfiguration)
+        fileIdentityStorage.saveUserId("user_id")
+        fileIdentityStorage.saveDeviceId("device_id")
+        val savedIdentity = fileIdentityStorage.load()
+        val expectedIdentity = Identity("user_id", "device_id")
+        assertEquals(savedIdentity, expectedIdentity)
+    }
+
+    @Test
+    fun `test FileIdentityStorage load from file success`() {
```
justin-fiedler, line 27: nit: I think it would be good to isolate this test to do its own set up rather than being dependent on the previous test.

justin-fiedler (approved):

id/src/test/kotlin/com/amplitude/id/FileIdentityStorageTest.kt
qingzhuozhen: Yeah, I tried to use some other method earlier though didn't work well. Let me just extract the setup to separate method.

PR #28, opened 2022-04-23
feat: add unit tests in event bridge module

qingzhuozhen:
### Summary

- add unit tests in event bridge module

### Checklist

### DIFF
--- a/event-bridge/src/test/kotlin/com/amplitude/eventbridge/EventBridgeTest.kt
+++ b/event-bridge/src/test/kotlin/com/amplitude/eventbridge/EventBridgeTest.kt
@@ -1,6 +1,7 @@
 package com.amplitude.eventbridge
 
 import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.TestInstance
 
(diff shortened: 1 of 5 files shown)


event-bridge/src/test/kotlin/com/amplitude/eventbridge/EventBridgeTest.kt
```diff
@@ -0,0 +1,55 @@
...
+        eventBridge.setEventReceiver(
+            EventChannel.IDENTIFY,
+            object : EventReceiver {
+                override fun receive(channel: EventChannel, event: Event) {
+                    assertEquals(event, identifyEvent)
+                }
+            }
+        )
+        eventBridge.sendEvent(EventChannel.EVENT, testEvent)
+        eventBridge.sendEvent(EventChannel.IDENTIFY, identifyEvent)
```
justin-fiedler, line 32: This could pass if the EventReceiver never gets called. Maybe add additional checks to be safe.

```
var wasEventRecieverCalled = false;
eventBridge.setEventReceiver(EventChannel.EVENT, object : EventReceiver {
      override fun receive(channel: EventChannel, event: Event) {
          assertEquals(event, testEvent)
          wasEventRecieverCalled = true;
      }
})

....

assertTrue(wasEventRecieverCalled)
```

justin-fiedler (approved):

event-bridge/src/test/kotlin/com/amplitude/eventbridge/EventBridgeTest.kt
qingzhuozhen: I guess the assertEquals check inside receiver can work? Let me add that to safeguard this.

PR #33, opened 2022-05-13
feat: expose more api for support

yuhao900914:
### Summary

Expose more APIs for identify and group identify call, in order to be easily used in ampli kotlin v2 SDK. 

### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/events/Identify.kt
+++ b/core/src/main/java/com/amplitude/core/events/Identify.kt
@@ -541,6 +541,10 @@ class Identify() {
         return this
     }
 
+    fun setUserProperty(property: String, value: Any) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/amplitude/core/events/Identify.kt
```diff
@@ -541,6 +541,10 @@ class Identify() {
         return this
     }
 
+    fun setUserProperty(property: String, value: Any) {
```
qingzhuozhen, line 544: Would it make more sense to be just set method?

qingzhuozhen (commented):
can you help fix the lint? Ktlint style is checked in ci action

qingzhuozhen (approved):

PR #35, opened 2022-06-01
feat: add configuration java support

qingzhuozhen:
### Summary

- update configuration to support better in java
- add example to init optionally with configuration

### Checklist


core/src/main/java/com/amplitude/core/Configuration.kt
```diff
@@ -8,21 +8,21 @@ import com.amplitude.core.utilities.InMemoryStorageProvider
 typealias EventCallBack = (BaseEvent, status: Int, message: String) -> Unit
 
 open class Configuration @JvmOverloads constructor(
-    val apiKey: String,
-    var flushQueueSize: Int = FLUSH_QUEUE_SIZE,
-    var flushIntervalMillis: Int = FLUSH_INTERVAL_MILLIS,
-    var instanceName: String = DEFAULT_INSTANCE,
-    var optOut: Boolean = false,
-    val storageProvider: StorageProvider = InMemoryStorageProvider(),
```
yuhao900914, line 16: Should this be va r or val? do we allowed to override?

qingzhuozhen: Good point, I was wondering about this as well. That time I changed to var to enable it to be set, taking another thought, maybe we can use this as a val first until it is need in the future. Let me change it back to val for now.

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -1,32 +1,32 @@
...
 import com.amplitude.core.Configuration
 import com.amplitude.core.EventCallBack
 import com.amplitude.core.LoggerProvider
 import com.amplitude.core.ServerZone
 import com.amplitude.core.StorageProvider
+import com.amplitude.core.events.Plan
 
 class Configuration @JvmOverloads constructor(
-    apiKey: String,
+    override val apiKey: String,
```
yuhao900914, line 14: why we use override here for api key?

qingzhuozhen: Good catch, that was accidentally added..

PR #37, opened 2022-06-03
feat: set the client userId and deviceId in identify call

yuhao900914:
### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/Amplitude.kt
+++ b/core/src/main/java/com/amplitude/core/Amplitude.kt
@@ -130,7 +130,7 @@ open class Amplitude internal constructor(
      * @return the Amplitude instance
      */
     @JvmOverloads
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -130,7 +130,7 @@ open class Amplitude internal constructor(
      * @return the Amplitude instance
      */
     @JvmOverloads
-    fun identify(userProperties: Map<String, Any>, options: EventOptions? = null): Amplitude {
+    fun identify(userProperties: Map<String, Any?>?, options: EventOptions? = null): Amplitude {
```
yuhao900914, line 133: This is for enable ampli identify call to call core identify. 
Screen Shot 2022-06-03 at 12 30 28 PM
The properties type is Map<String, Any?>? but here required Map<String, Any>. So updated this and add more logic to handle the null.

qingzhuozhen (approved):

PR #39, opened 2022-06-18
chore: setup common-android for publishing

bgiori:
### Checklist

* [X] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-Kotlin/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?:   no

### DIFF
--- a/common-android/build.gradle
+++ b/common-android/build.gradle
@@ -4,6 +4,17 @@ plugins {
     id 'de.mannodermaus.android-junit5'
 }
 
+ext {
+    PUBLISH_NAME = 'Amplitude Android Kotlin Common Code'
+    PUBLISH_DESCRIPTION = 'Amplitude Kotlin client-side common code for Android'
+    PUBLISH_GROUP_ID = 'com.amplitude'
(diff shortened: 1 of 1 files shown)


common-android/build.gradle
```diff
@@ -39,9 +50,6 @@ android {
 
 dependencies {
     api project(":common")
-    implementation 'androidx.core:core-ktx:1.7.0'
-    implementation 'androidx.appcompat:appcompat:1.4.1'
-    implementation 'com.google.android.material:material:1.5.0'
```
bgiori, line 44: Not sure what these were used for so I removed them, lmk if I should add them back.

qingzhuozhen: I guess these were copied earlier from android, we can remove this if this won't hurt build. Do you still want to have an github for this or manual for now?

qingzhuozhen (approved):

common-android/build.gradle
bgiori: I'll add an action

PR #40, opened 2022-06-24
feat: add remnant events migration plugin

liuyang1520:
### Summary
This PR adds a remnant events migration plugin to send the remnant events in Sqlite generated by old Android Java SDK, including:
- DB utils for retrieving the entries in sqlite database
- Covert the event to new format
- Send the event with this new Kotlin SDK

Todos:
- [x] Remove events from sqlite after tracking with new SDK
- [x] Add unit tests
- [x] Create a confluence page for how to test this: https://amplitude.atlassian.net/wiki/spaces/GOV/pages/94437526/Android#How-to-test-both-old-and-new-Android-SDK-together

### Checklist
yes
no


android/src/main/java/com/amplitude/android/utilities/DatabaseStorage.kt
```diff
@@ -0,0 +1,214 @@
...
+    const val EVENT_TABLE_NAME = "events"
+    const val ID_FIELD = "id"
+    const val EVENT_FIELD = "event"
+}
+
+/**
+ * The SDK doesn't need to write/read from local sqlite database.
+ * This storage class is used for migrating events only.
+ */
+class DatabaseStorage(context: Context) : Storage, SQLiteOpenHelper(
```
qingzhuozhen, line 38: Do we need the Storage interface here?

android/src/main/java/com/amplitude/android/plugins/RemnantEventsMigrationPlugin.kt
```diff
@@ -0,0 +1,142 @@
...
+        super.setup(amplitude)
+        val configuration = amplitude.configuration as Configuration
+        contextProvider =
+            AndroidContextProvider(configuration.context, configuration.locationListening)
+
+        val databaseStorage = DatabaseStorageProvider().getStorage(amplitude)
+        var maxMigratedEventId: Long = 0
+        try {
+            @Suppress("UNCHECKED_CAST")
+            val remnantEvents = databaseStorage.readEventsContent() as List<JSONObject>
```
qingzhuozhen, line 40: This is for events rights? Do you cover the identify events, and user_id, device_id, etc.?

android/build.gradle
```diff
@@ -55,15 +55,15 @@ dependencies {
     embed project(':common')
     embed project(':common-jvm')
     embed project(':common-android')
-    embed project(':core')
+    api project(':core')
```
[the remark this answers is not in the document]
liuyang1520: Thanks for reviewing! I have to change this to `api` as the unit tests needs to refer to some classes in `core`. I think `core` has similar issue: https://github.com/amplitude/Amplitude-Kotlin/blob/main/core/build.gradle#L22-L25. Let me know if you have other methods to handle this.

android/src/main/java/com/amplitude/android/plugins/RemnantEventsMigrationPlugin.kt
```diff
@@ -0,0 +1,142 @@
...
+            if (maxMigratedEventId > 0) {
+                databaseStorage.removeEvents(maxMigratedEventId)
+            }
+        }
+    }
+}
+
+// Copy from core/src/main/java/com/amplitude/core/utilities/JSON.kt with minor changes.
+// To avoid unexpected results if open this extension to wide permission.
+internal fun JSONObject.toBaseEvent(): BaseEvent {
```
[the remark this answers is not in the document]
liuyang1520: Two problems here to reuse:
1. Those extension functions are `internal`, see: https://github.com/amplitude/Amplitude-Kotlin/blob/main/core/src/main/java/com/amplitude/core/utilities/JSONUtil.kt#L159. Some are even too wide to be private: https://github.com/amplitude/Amplitude-Kotlin/blob/0840b5d5fc538b2f088538365586cbf5babb819d/core/src/main/java/com/amplitude/core/utilities/JSON.kt#L59. If we want to reuse, we need to extend all `JSONObject` across all modules, unexpected behavior might happen.
2. There are diffs IMO, for example: https://github.com/amplitude/Amplitude-Kotlin/blob/2bfb684d9ad760d7e544f9791be4e232d4cf33aa/android/src/main/java/com/amplitude/android/plugins/RemnantEventsMigrationPlugin.kt#L102. I am not too sure on this part. Let me know if I miss anything.

android/src/main/java/com/amplitude/android/utilities/DatabaseStorage.kt
liuyang1520: It doesn't have to be, but my understanding is since we already have this interface, we can use it as a standard `Storage` interface, providing similar code like `provider` and methods, easier to integrate, and potentially reuse if database access is needed in the future. Most of the methods are not implemented at the moment. So far I don't see issues with implementing this interface. Let me know if you have any concerns!

android/src/main/java/com/amplitude/android/plugins/RemnantEventsMigrationPlugin.kt
liuyang1520: Yes, events, example:
```json
{
    "event_type": "marvin-test",
    "timestamp": 1656022776834,
    "user_id": "marvin",
    "device_id": "d34d76b7-ff6a-40ae-8dd3-fc93ece8fdceR",
    "session_id": 1656018188962,
    "uuid": "70b13b79-8530-4399-b39e-1aaf03ce9734",
    "sequence_number": 8,
    "version_name": "1.0",
    "os_name": "android",
    "os_version": "13",
    "api_level": 33,
    "device_brand": "google",
    "device_manufacturer": "Google",
    "device_model": "sdk_gphone64_arm64",
    "carrier": "T-Mobile",
    "country": "US",
    "language": "en",
    "platform": "Android",
    "library": {
        "name": "amplitude-android",
        "version": "2.37.1-SNAPSHOT"
    },
    "api_properties": {
        "limit_ad_tracking": false,
        "gps_enabled": true
    },
    "event_properties": {},
    "user_properties": {},
    "groups": {},
    "group_properties": {}
}
```
`user_id` and `device_id` are already embedded. I am not sure about `identify events` you mentioned, are they special events? (I didn't find any related desc in https://developers.amplitude.com/docs/android, or in my local db). Do you know how to generate those if they are not the events I posted above? Once I figure out how to generate it and confirm we need it, I can update to try to cover those too.

android/build.gradle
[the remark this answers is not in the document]
liuyang1520: `testImplementation project(:core)`

android/src/main/java/com/amplitude/android/plugins/RemnantEventsMigrationPlugin.kt
liuyang1520: - get `user_id`, `device_id`, `session_id`, `opt_out`...
- get `identify` events

falconandy:
1. Migration logic is changed to write legacy data directly to new storages.
2. Migration logic is moved from `Plugin` to let run it before any plugins including standard plugin (AndroidContextPlugin, etc). I called it `Initializer` but maybe simple configuration flag like `runLegacyDataMigration` is enough.  Reason: device/user id logic is pretty complex (classes `State`, `IdentityContainer`, `IdentityListener`, etc) and asynchronous - AndroidContextPlugin concurrently sets random device id on first run if device id is currently empty - sometimes the random value wins, sometimes - value from legacy data wins (if migration logic is implemented in a plugin executes after AndroidContextPlugin).
3. Supported: on first run since upgrade (recommended) - device/user id, session id, last event date/id, events, identifies, intercepted identifies. If migration was enabled later after upgrade - only events are migrated.
4. Supported sqlite db versions - v3 (2015 year) and current v4 (2/9/23, added table `identify_interceptor`)
5. Added support for instance names.
6. Some legacy fields are converted to new names/format: `library` -> from object to string, `timestamp` -> `time`, `adid`, `android_app_set_id`.
7. Added methods `getUserId()`/ `getDeviceId` to android `Amplitude`
8. After migration legacy sqlite data are cleaned except `deviceId`/`userId` - maybe they will be required later to link legacy/new ids, for example.
9. Added tests for v3/v4 sqlite databases - currently cover `on first run since upgrade` case

justin-fiedler (commented):
~~Thanks @falconandy. I see what you mean about the complexity of the ID logic. I'm trying to think if there is anything else we can do to prevent adding the new concept of "initializers", but nothing is coming to mind.~~

~~Any ideas @qingzhuozhen or @liuyang1520?~~

android/src/test/java/com/amplitude/android/AmplitudeSessionTest.kt
```diff
@@ -54,6 +55,11 @@ class AmplitudeSessionTest {
         IdentityContainer.getInstance(configuration)
     }
 
+    @AfterEach
+    fun tearDown() {
+        Thread.sleep(200)
```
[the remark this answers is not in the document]
falconandy: I added the delay to let complete asynchronous handlers:

```
class AndroidContextPlugin : Plugin {
    ...
    override fun setup(amplitude: Amplitude) {
        ...
        initializeDeviceId(configuration)
    }

    fun initializeDeviceId(configuration: Configuration) {
        ...
        amplitude.setDeviceId(randomId)
    }
``` 

```
open class Amplitude internal constructor(
...
    fun setDeviceId(deviceId: String): Amplitude {
            isBuilt.await()
            idContainer.identityManager.editIdentity().setDeviceId(deviceId).commit()
        }
        return this
    }
```

`idContainer.identityManager.editIdentity().setDeviceId(deviceId).commit()` saves updated identify data to storage. If file storage is used (by default) temp directory can be removed already and the saving fails.

I've removed the delay in tests and switched to in-memory identify storage.

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -58,7 +59,15 @@ open class Amplitude internal constructor(
         require(configuration.isValid()) { "invalid configuration" }
         timeline = this.createTimeline()
         logger = configuration.loggerProvider.getLogger(this)
-        idContainer = createIdentityContainer()
+        storage = configuration.storageProvider.getStorage(this)
```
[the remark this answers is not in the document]
falconandy: I think, it is safe.
Maybe the Kotlin SDK contains too much suspend/coroutines logic (in comparison with Amplitude-Swift) and can be simplified.

android/src/main/java/com/amplitude/android/migration/RemnantDataMigration.kt
```diff
@@ -0,0 +1,178 @@
...
+
+    lateinit var amplitude: Amplitude
+    lateinit var databaseStorage: DatabaseStorage
+
+    override fun execute(amplitude: Amplitude) {
+        this.amplitude = amplitude
+        runBlocking {
+            databaseStorage = DatabaseStorageProvider().getStorage(amplitude)
+
+            val firstRunSinceUpgrade = amplitude.storage.read(Storage.Constants.LAST_EVENT_TIME)?.toLongOrNull() == null
```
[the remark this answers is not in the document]
falconandy: Now `databaseStorage` is used in any case and should be created.

Supported cases:
1. android-sdk -> **kotlin-sdk (with migration)**. 
2. android-sdk -> kotlin-sdk (without migration support) -> **kotlin-sdk (with migration)**. 
3. android-sdk -> kotlin-sdk (with migration)  -> **kotlin-sdk (with migration**) -> ...

1: all data are migrated.
2: only events are migrated.
3: no data are migrated actually but migration still executes. If we want to skip migration we should explicitly save to current storage some flag like `migration was executed` (vs `migration date`).

android/src/main/java/com/amplitude/android/migration/RemnantDataMigration.kt
```diff
@@ -0,0 +1,178 @@
...
+        }
+    }
+
+    private suspend fun moveEvents() {
+        try {
+            val remnantEvents = databaseStorage.readEventsContent()
+
+            for (event in remnantEvents) {
+                val rowId = moveEvent(event, amplitude.storage)
+                databaseStorage.removeEvent(rowId)
```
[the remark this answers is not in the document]
falconandy: Fixed

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -39,8 +42,10 @@ open class Configuration @JvmOverloads constructor(
     var minTimeBetweenSessionsMillis: Long = MIN_TIME_BETWEEN_SESSIONS_MILLIS,
     var trackingSessionEvents: Boolean = true,
     override var identifyBatchIntervalMillis: Long = IDENTIFY_BATCH_INTERVAL_MILLIS,
-    override var identifyInterceptStorageProvider: StorageProvider = AndroidStorageProvider()
-) : Configuration(apiKey, flushQueueSize, flushIntervalMillis, instanceName, optOut, storageProvider, loggerProvider, minIdLength, partnerId, callback, flushMaxRetries, useBatch, serverZone, serverUrl, plan, ingestionMetadata, identifyBatchIntervalMillis, identifyInterceptStorageProvider) {
+    override var identifyInterceptStorageProvider: StorageProvider = AndroidStorageProvider(),
+    override var identityStorageProvider: IdentityStorageProvider = FileIdentityStorageProvider(),
+    initializers: List<Initializer>? = null
```
justin-fiedler, line 47: > Thanks @falconandy. I see what you mean about the complexity of the ID logic. I'm trying to think if there is anything else we can do to prevent adding the new concept of "initializers", but nothing is coming to mind.

Given the complexities with id, and now that Migration code/plugin is a part of the SDK codebase, I think instead of adding the new concept of `initializers` we can instead include the migration logic by default and toggle it on/off via Configuration. In this case the migration logic no longer needs to use the Plugin interface.

Ideally if `migrateLegacyData` is `false` (the default) no SQLLite code is run at all.

```
val amplitude = Amplitude(Configuration(migrateLegacyData: true));
```

@falconandy can you please refactor accordingly. Thank you 🙏

justin-fiedler (changes requested):

android/src/main/java/com/amplitude/android/Configuration.kt
falconandy: Refactored, added the `migrateLegacyData` flag (default is `false`)

core/src/main/java/com/amplitude/core/Amplitude.kt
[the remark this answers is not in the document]
falconandy: Refactored: moved back to `build()`

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -39,8 +41,10 @@ open class Configuration @JvmOverloads constructor(
     var minTimeBetweenSessionsMillis: Long = MIN_TIME_BETWEEN_SESSIONS_MILLIS,
     var trackingSessionEvents: Boolean = true,
     override var identifyBatchIntervalMillis: Long = IDENTIFY_BATCH_INTERVAL_MILLIS,
-    override var identifyInterceptStorageProvider: StorageProvider = AndroidStorageProvider()
-) : Configuration(apiKey, flushQueueSize, flushIntervalMillis, instanceName, optOut, storageProvider, loggerProvider, minIdLength, partnerId, callback, flushMaxRetries, useBatch, serverZone, serverUrl, plan, ingestionMetadata, identifyBatchIntervalMillis, identifyInterceptStorageProvider) {
+    override var identifyInterceptStorageProvider: StorageProvider = AndroidStorageProvider(),
+    override var identityStorageProvider: IdentityStorageProvider = FileIdentityStorageProvider(),
+    val migrateLegacyData: Boolean = false,
+) : Configuration(apiKey, flushQueueSize, flushIntervalMillis, instanceName, optOut, storageProvider, loggerProvider, minIdLength, partnerId, callback, flushMaxRetries, useBatch, serverZone, serverUrl, plan, ingestionMetadata, identifyBatchIntervalMillis, identifyInterceptStorageProvider, identityStorageProvider) {
```
yuhao900914, line 47: Are we providing the ability to overwrite the identityStorageProvider here?

samples/kotlin-android-app/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
```
yuhao900914, line 8: Question: Do we need this permission for all Android 6.0 (Marshmallow) and higher like what we mentioned [here](https://www.docs.developers.amplitude.com/data/sdks/android/#add-dependencies)?  Or this is related to the migration logic?

android/src/test/java/com/amplitude/android/migration/RemnantDataMigrationTest.kt
```diff
@@ -0,0 +1,168 @@
...
+            )
+        )
+
+        // Check migrated data after RemnantEventsMigrationPlugin
+        runBlocking {
+            amplitude.isBuilt.await()
+
+            val identity = amplitude.idContainer.identityManager.getIdentity()
+            if (inputStream != null && migrateLegacyData) {
+                Assertions.assertEquals("22833898-c487-4536-b213-40f207abdce0R", identity.deviceId)
```
yuhao900914, line 65: Question: where are the original legacy data come from?  Can we use some shared test Constance instead of using real value here?

android/src/main/java/com/amplitude/android/migration/RemnantDataMigration.kt
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    lateinit var databaseStorage: DatabaseStorage
+
+    suspend fun execute() {
+        databaseStorage = DatabaseStorageProvider.getStorage(amplitude)
+
+        val firstRunSinceUpgrade = amplitude.storage.read(Storage.Constants.LAST_EVENT_TIME)?.toLongOrNull() == null
+
+        moveDeviceAndUserId()
```
liuyang1520, line 36: If any of these `move` method fails, this continues running, as the `catch` block will catch any exception in each method. As there is no "transaction" here, will this cause any data inconsistency issue. I think it is probably fine to leave like this, just to double check.

android/src/main/java/com/amplitude/android/Configuration.kt
falconandy: yes, similar to `storageProvider` and `identifyInterceptStorageProvider`

samples/kotlin-android-app/src/main/AndroidManifest.xml
falconandy: Was added in initial commit 'feat: add remnant events migration plugin Marvin Liu 6/24/22, 11:27 AM'
Looks like it can be removed - I run the sample application without it (Android emulator).
Removed.

android/src/test/java/com/amplitude/android/migration/RemnantDataMigrationTest.kt
falconandy: Yes, it comes from `android/src/test/resources/legacy_v*.sqlite` files
I copied the files from my Android emulator file system.
The identifiers are unique random uids, the files do not contain api keys.

android/src/main/java/com/amplitude/android/migration/RemnantDataMigration.kt
falconandy: The `move` methods are independent of each other.

`moveDeviceAndUserId`/ `moveSessionData` are safe - they overwrite current value only if the current value is not set.
`moveInterceptedIdentifies`/ `moveIdentifies` run once only - an event are skipped on exception
`moveEvents` can run several times for en event (say, the event is written to new storage but removing from legacy storage fails for some reason). `insert_id` should prevent event duplication in this case

It looks like legacy field name is `uuid`, new name is `insert_id` - I've added conversion `uuid` -> `insert_id`

core/src/main/java/com/amplitude/core/Storage.kt
```diff
@@ -18,6 +19,8 @@ interface Storage {
 
     suspend fun writeEvent(event: BaseEvent)
 
+    suspend fun writeEvent(event: JSONObject)
```
justin-fiedler, line 22: @falconandy I'm guessing this was added to prevent double conversion of events during migration (JSON -> BaseEvent - JSON) but am a little concerned about adding JSONObject to the core Storage interface.

For example, JRE doesn't include JSON in the core libraries and it would need to be added as an additional dependency. The goal for this repo in the long run is to support both Android and JRE and this could complicate things.

Since the migration plugin (hopefully) only runs once, I think I would prefer the double conversion over adjusting the core Storage interface.

wdyt?

cc @qingzhuozhen @liuyang1520

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -70,29 +77,45 @@ open class Amplitude internal constructor(
...
+    protected fun createIdentityContainer(identityConfiguration: IdentityConfiguration) {
+        idContainer = IdentityContainer.getInstance(identityConfiguration)
         val listener = AnalyticsIdentityListener(store)
         idContainer.identityManager.addIdentityListener(listener)
         if (idContainer.identityManager.isInitialized()) {
             listener.onIdentityChanged(idContainer.identityManager.getIdentity(), IdentityUpdateType.Initialized)
         }
+    }
+
+    private fun build(): Deferred<Boolean> {
```
justin-fiedler, line 98: This used to be `open` maybe should we keep it that way for backwards compatiblity?

core/src/main/java/com/amplitude/core/Storage.kt
falconandy: @justin-fiedler Currently `core` package already depends on `org.json.*` types (JSONException, JSONObject, JSONArray).
Adding JSONObject to the core Storage interface doesn't change anything - JRE should add JSON as an additional dependency in any case.
wdyt?

core/src/main/java/com/amplitude/core/Amplitude.kt
falconandy: `build()` is invoked in Amplitude constructor (`init`) and should not be invoked explicitly. Actually I had a error in tests when I used `build().await()` instead of `isBuilt.await()`
If someone currently invokes `build()` explicitly - it is definitely a bug, I believe.

justin-fiedler: I was thinking more about overriding versus direct access to the method. What if we can make it `open protected` to allow overriding in subclasses but still prevent direct access?

falconandy: Fixed. `protected open` now

core/src/main/java/com/amplitude/core/Storage.kt
justin-fiedler: Hmm, got it. In that cause it matters less, but it still seems "less clean" to have to support JSON in addition and BaseEvent to define a storage.

I still would prefer double conversion over adding JSON to Storage interface. Let me know if you strongly disagree, thanks.

falconandy: Possible issue is differences in json encoding between legacy/current SDK. I added code to fix some ones (`timestamp` -> `time`, `uuid` -> `insert_id`, `library`) but maybe `legacy JSON` ->`event` -> `current JSON` looks/works better

justin-fiedler: Thanks, I saw the conversion code. I think you are handling the most important fields.

> legacy JSON ->event -> current JSON

This sounds good 👍 if we find any non-transferred fields in the future we can update the conversion code as needed.

justin-fiedler (approved):

core/src/main/java/com/amplitude/core/utilities/JSONUtil.kt
```diff
@@ -76,7 +76,7 @@ object JSONUtil {
         return eventsArray.toString()
     }
 
-    private fun truncate(obj: JSONObject?): JSONObject? {
+    private fun truncate(obj: JSONObject?): JSONObject {
```
justin-fiedler, line 79: Why was this updated? I feel in some cases we want to have nil for empty fields we don't want to transfer empty fields.

I looked at the checks and noticed new warnings in the build from this update. Can you please look into the warnings and make sure this won't cause any problems.

image

Last build before changes:
https://github.com/amplitude/Amplitude-Kotlin/actions/runs/4991994545/jobs/8939218744

After:
https://github.com/amplitude/Amplitude-Kotlin/actions/runs/4961668962/jobs/8878750018

justin-fiedler (changes requested):

core/src/main/java/com/amplitude/core/utilities/JSONUtil.kt
falconandy: > Why was this updated? I feel in some cases we want to have nil for empty fields we don't want to transfer empty fields.

Android Studio shows warning:
image

falconandy: > I looked at the checks and noticed new warnings in the build from this update. Can you please look into the warnings and make sure this won't cause any problems.

I tried locally this branch and `main` - both have the same build warnings after `clean`:
```
./gradlew clean
./gradlew build
```

I can fix some of them that look safe.

falconandy: Fixed some "simple" warnings

justin-fiedler: Thanks @falconandy. The build output looks much better 🙏

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -205,6 +237,15 @@ open class Amplitude internal constructor(
         return this
     }
 
+    /**
+     * Get the device id.
+     *
+     * @return Device id.
+     */
+    fun getDeviceId(): String? {
```
justin-fiedler, line 245: Note for later: Could the logic in `setDeviceId()` cause a bug in setting the deviceId? Should it check the return of await? E.g.
```
if (isBuilt.await()) {
idContainer.identityManager.editIdentity().setDeviceId(deviceId).commit()
}
```

falconandy: @justin-fiedler Currently `build()` (and `isBuilt.await()`) always returns `true`
https://github.com/amplitude/Amplitude-Kotlin/blob/main/core/src/main/java/com/amplitude/core/Amplitude.kt#L108

PR #42, opened 2022-06-26
fix: fix group call

qingzhuozhen:
### Summary

- update group call from use identify to use groups directly

### Checklist


kevinpagtakhan (approved):
🎉

PR #48, opened 2022-07-22
feat: add reset method to reset userId and deviceId

liuyang1520:
### Summary
This PR adds the `#reset` method, which does the following things:
- `setUserId` to null
- `setDeviceId` to new generated one

### Checklist

No

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -24,6 +24,7 @@ open class Amplitude(
     internal var lastEventId: Long = 0
     var lastEventTime: Long = -1
     private var previousSessionId: Long = -1
+    private lateinit var androidContextPlugin: AndroidContextPlugin
 
     override fun build() {
         val storageDirectory = (configuration as Configuration).context.getDir("${FileStorage.STORAGE_PREFIX}-${configuration.instanceName}", Context.MODE_PRIVATE)
@@ -53,10 +54,26 @@ open class Amplitude(
             lastEventTime = storage.read(Storage.Constants.LAST_EVENT_TIME) ?. let {
                 it.toLong()
             } ?: -1
-            add(AndroidContextPlugin())
+            androidContextPlugin = AndroidContextPlugin()
+            add(androidContextPlugin)
             add(AndroidLifecyclePlugin())
         }
         add(AmplitudeDestination())
+    }
+
+    /**
+     * Reset identity:
+     *  - reset userId to "null"
+     *  - reset deviceId via AndroidContextPlugin
+     * @return the Amplitude instance
+     */
+    override fun reset(): Amplitude {
+        this.setUserId(null)
+        amplitudeScope.launch(amplitudeDispatcher) {
+            idContainer.identityManager.editIdentity().setDeviceId(null).commit()
+            androidContextPlugin.initializeDeviceId(configuration as Configuration)
+        }
+        return this
     }
 
     fun onEnterForeground(timestamp: Long) {
--- a/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
+++ b/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
@@ -30,7 +30,7 @@ class AndroidContextPlugin : Plugin {
         return event
     }
 
-    private fun initializeDeviceId(configuration: Configuration) {
+    fun initializeDeviceId(configuration: Configuration) {
         val deviceId = amplitude.store.deviceId
         if (deviceId != null && validDeviceId(deviceId) && !deviceId.endsWith("S")) {
             return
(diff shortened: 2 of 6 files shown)


android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
```diff
@@ -30,7 +30,7 @@ class AndroidContextPlugin : Plugin {
         return event
     }
 
-    private fun initializeDeviceId(configuration: Configuration) {
+    fun initializeDeviceId(configuration: Configuration) {
```
yuhao900914, line 33: Question: Should this be protected?

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -183,6 +184,18 @@ open class Amplitude internal constructor(
         return this
     }
 
+    /**
```
yuhao900914, line 187: Do we need `reset` in core?

android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
liuyang1520: Thanks for reviewing! I think it is the same, tried adding `protected`, IDE shows `'protected' visibility is effectively 'private' in a final class`.

core/src/main/java/com/amplitude/core/Amplitude.kt
liuyang1520: I think this is an open question. @qingzhuozhen feel free to suggest more. My understanding is `core` provides a basic implementation for all functions without the environment specific context. Let me know if this is wrong or you have any concerns!

qingzhuozhen (approved):

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -53,12 +54,28 @@ open class Amplitude(
...
+    /**
+     * Reset identity:
+     *  - reset userId to "null"
+     *  - reset deviceId via AndroidContextPlugin
+     * @return the Amplitude instance
+     */
+    override fun reset(): Amplitude {
+        this.setUserId(null)
+        amplitudeScope.launch(amplitudeDispatcher) {
+            idContainer.identityManager.editIdentity().setDeviceId(null).commit()
```
qingzhuozhen, line 73: Do we need this?

android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
qingzhuozhen: internal?

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -53,12 +54,28 @@ open class Amplitude(
...
+            androidContextPlugin.initializeDeviceId(configuration as Configuration)
```
qingzhuozhen, line 74: A good thought for if we still respect the settings. I wonder if in this case we can just call the uuid method and skip the logic to check adid or appset id

core/src/main/java/com/amplitude/core/Amplitude.kt
qingzhuozhen: That is a good question. I guess we could keep this in core if we have the user id and device id method. @kevinpagtakhan @justin-fiedler any thoughts?

If we have the method in core, do we still need the android to have a different version?

yuhao900914: Feels in the core we cannot get the android context? 
initializeDeviceId method is only in androdiContextPlugin now. there is no such method in setup in the ContextPlugin

liuyang1520: Yes, core doesn't have the Android context. I don't want to introduce the Android dependency for core, imaging the code is running inside a Java server. Two ways list below and I pick up the min change way (second):
- Introduce a new plugin (e.g., ContextPlugin) with reset method, then core can use timeline to call all reset method of these plugins
- Core has a basic implementation, different context might override depending on runtime env

android/src/main/java/com/amplitude/android/Amplitude.kt
liuyang1520: Yes, since `initializeDeviceId` will skip if `deviceId` already exists

android/src/main/java/com/amplitude/android/Amplitude.kt
liuyang1520: So we just set the deviceId to uuid without respecting the settings? I am ok with both, but not sure whether this will introduce confusion to users.

qingzhuozhen: yeah, I guess that could work. We can also keep the current way and see how it goes

qingzhuozhen (approved):

PR #50, opened 2022-07-27
fix: invalid json array string for android file storage

bohan-amplitude:
### Summary

fix unterminated array
fix expected literal value at character 0

### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
+++ b/core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
@@ -36,13 +36,10 @@ class EventsFileManager(
      * stores the event
      */
     suspend fun storeEvent(event: String) = mutex.withLock {
-        var newFile = false
         var file = currentFile()
         if (!file.exists()) {
             // create it
             file.createNewFile()
-            start(file)
-            newFile = true
         }
 
         // check if file is at capacity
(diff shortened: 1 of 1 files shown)


liuyang1520 (approved):
LGTM, to double check, is this to handle the case that when a file is not `newFile`, but doesn't have any content inside?

bohan-amplitude:
@liuyang1520 I found two cases in the issue report, one is file already exist but empty, another is missing comma between start_session and end_session events. Still don't know what really caused the second one. But by check file length I think we can cover both cases.

qingzhuozhen (approved):

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -51,12 +48,12 @@ class EventsFileManager(
             // update index
             file = currentFile()
             file.createNewFile()
-            start(file)
-            newFile = true
         }
 
         var contents = ""
-        if (!newFile) {
```
yuhao900914, line 59: Question: Will remove the newFile flag have a concurrency issue?

bohan-amplitude: I think there is concurrency issue somewhere else since the example of customer's payload have string like [[{event}{event} that are caused by two threads writing to file. But the newFile is a variable inside a function scope and not shared by multiple threads, it's not related to concurrency issue.

yuhao900914 (approved):

PR #55, opened 2022-08-03
fix: Amplitude.build() is not main thread safe on Android

bohan-amplitude:
### Summary

put IO in coroutines for Amplitude#build() on Android

### Checklist

* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-Kotlin/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?no

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -27,33 +27,27 @@ open class Amplitude(
     private lateinit var androidContextPlugin: AndroidContextPlugin
 
     override fun build() {
-        val storageDirectory = (configuration as Configuration).context.getDir("${FileStorage.STORAGE_PREFIX}-${configuration.instanceName}", Context.MODE_PRIVATE)
-        idContainer = IdentityContainer.getInstance(
-            IdentityConfiguration(
-                instanceName = configuration.instanceName,
-                apiKey = configuration.apiKey,
-                identityStorageProvider = FileIdentityStorageProvider(),
-                storageDirectory = storageDirectory
-            )
-        )
-        val listener = AnalyticsIdentityListener(store)
-        idContainer.identityManager.addIdentityListener(listener)
-        if (idContainer.identityManager.isInitialized()) {
-            listener.onIdentityChanged(idContainer.identityManager.getIdentity(), IdentityUpdateType.Initialized)
-        }
         amplitudeScope.launch(amplitudeDispatcher) {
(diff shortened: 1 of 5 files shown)


bohan-amplitude:
@qingzhuozhen I moved the disk IO into the coroutines for Amplitude#build() based on the customer issue: If called Amplitude#build() on main thread on Android. The Android strict mode will detect IO on main thread and throw DiskReadViolation. Want to check if this will case issue since Amplitude#build() is possible to return before idContainer init

bohan-amplitude:
@liuyang1520 Could you help check the reset id test

liuyang1520:
> @liuyang1520 Could you help check the reset id test

Forked this branch locally, moved the `idContainer` init outside `amplitudeScope`, test is passing for me. Trying to understand the case here, is it possible to init `idContainer` outside `amplitudeScope`? 

```kotlin
        val storageDirectory = (configuration as Configuration).context.getDir("${FileStorage.STORAGE_PREFIX}-${configuration.instanceName}", Context.MODE_PRIVATE)
        idContainer = IdentityContainer.getInstance(
            IdentityConfiguration(
                instanceName = configuration.instanceName,
                apiKey = configuration.apiKey,
                identityStorageProvider = FileIdentityStorageProvider(),
                storageDirectory = storageDirectory
            )
        )
...
```

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -27,13 +27,11 @@ open class Amplitude(
     private lateinit var androidContextPlugin: AndroidContextPlugin
 
     override fun build() {
-        val storageDirectory = (configuration as Configuration).context.getDir("${FileStorage.STORAGE_PREFIX}-${configuration.instanceName}", Context.MODE_PRIVATE)
         idContainer = IdentityContainer.getInstance(
             IdentityConfiguration(
                 instanceName = configuration.instanceName,
                 apiKey = configuration.apiKey,
-                identityStorageProvider = FileIdentityStorageProvider(),
-                storageDirectory = storageDirectory
```
qingzhuozhen, line 36: would the identity storage still create correctly in this case? The identity module will trigger the initialization right?

liuyang1520 (approved):
LGTM

android/src/main/java/com/amplitude/android/Amplitude.kt
bohan-amplitude: moved the storage init after set the directory path

PR #57, opened 2022-08-09
chore: use the github reusable workflow to create jira issue

yuhao900914:
### Checklist

### DIFF
--- a/.github/workflows/jira-issue-create.yml
+++ b/.github/workflows/jira-issue-create.yml
@@ -1,39 +1,21 @@
 # Creates jira tickets for new github issues to help triage
-name: Jira Issue Creator
+name: Jira Issue Creator For Kotlin
 
 on:
   issues:
     types: [opened]
+  workflow_call:
+    inputs:
+      label:
+        type: string
 
 jobs:
-  build:
-    runs-on: ubuntu-18.04
-    environment: Jira
-    name: SDK Bot Jira Issue Creation
-    steps:
-      - name: Login
-        uses: atlassian/gajira-login@master
-        env:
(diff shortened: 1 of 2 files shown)


liuyang1520 (approved):

.github/workflows/jira-issue-create.yml
```diff
@@ -1,39 +1,21 @@
...
-            "customfield_10006": 1391,
-            "customfield_11481": {"value": "Governance 🕵️"},
-            "customfield_11200": {"value": "Developer Experience"},
-            "labels": ["kotlin"]
-          }' # sprint, pillar, pod, labels
-
-      - name: Log created issue
-        run: echo "Issue AMP-${{ steps.create.outputs.issue }} was created"
+  call-workflow-passing-data:
+    uses: amplitude/Amplitude-TypeScript/.github/workflows/jira-issue-create-template.yml@main
```
qingzhuozhen, line 14: Do we have plan to use DXOC in the future for triage? Or we can keep this for now?

qingzhuozhen (approved):

.github/workflows/jira-issue-create.yml
yuhao900914: I think no matter if we want to use DXOC or not, with this reusable workflow, we only need to update the template located in amplitude-typescript. Don't need to update the workflow one by one.

yuhao900914: Oh, also need to update the project secret in each repo for now.

qingzhuozhen: yeah, just wonder if we need an extra field besides label. Maybe we can keep what we have for now and reuse it.

PR #61, opened 2022-08-11
fix: Unterminated arrays caused by multi client instances with same name and api key

bohan-amplitude:
### Summary

* Put mutex to class companion to lock for possible multi instance access
* Reduce duplicate events requests from same event file
* Put curFile reference to companion and managed by apiKey
* Handle invalid events json string: delete file and clean up callback map

### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
+++ b/core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
@@ -17,7 +17,6 @@ class EventsFileManager(
 ) {
     init {
         createDirectory(directory)
-        registerShutdownHook()
     }
 
     private val fileIndexKey = "amplitude.events.file.index.$apiKey"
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -15,39 +17,39 @@ class EventsFileManager(
...
-    private var os: FileOutputStream? = null
-
-    private var curFile: File? = null
-
-    private val mutex = Mutex()
-
     companion object {
         const val MAX_FILE_SIZE = 975_000 // 975KB
+        val writeMutex = Mutex()
+        val readMutex = Mutex()
```
liuyang1520, line 27: With the mutex separation, so now the file can be read and write at the same time?

liuyang1520 (approved):
LGTM

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
bohan-amplitude: We used to only lock on write and don't do restriction on read. I saw a lot of logs read duplicate files and send requests when debugging. So added a lock to to put filepath in a set then read file. The filepath in the set will block one read to reduce the duplicate requests. I don't block the read until delete the file is because sometime errors happened before processing response. The file will not be deleted and will not be read again.

liuyang1520: Got it, thanks for providing more details! Out of scope for this PR, just curious some files will remain on the device (untouched forever) if not deleted and not be read again, do we clean those files?

PR #63, opened 2022-09-01
feat: add ingestion_metadata field

liuyang1520:
### Summary
- feat: add ingestion_metadata field

Add this option in the configuration, which can be used in the library wrapper/code generation/dynamic loading cases, for setting the `ingestion_metadata` information.

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
+++ b/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
@@ -150,6 +150,11 @@ class AndroidContextPlugin : Plugin {
                 event.plan = it
             }
         }
+        event.ingestionMetadata ?: let {
+            amplitude.configuration.ingestionMetadata ?. let {
+                event.ingestionMetadata = it
+            }
+        }
     }
 
     private fun getAndroidAmplitude(): com.amplitude.android.Amplitude {
--- a/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
+++ b/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
@@ -147,12 +147,12 @@ class AndroidContextPlugin : Plugin {
         }
         event.plan ?: let {
             amplitude.configuration.plan ?. let {
-                event.plan = it
+                event.plan = it.clone()
             }
         }
         event.ingestionMetadata ?: let {
             amplitude.configuration.ingestionMetadata ?. let {
-                event.ingestionMetadata = it
+                event.ingestionMetadata = it.clone()
             }
         }
     }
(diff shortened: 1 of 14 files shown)


android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
```diff
@@ -150,6 +150,11 @@ class AndroidContextPlugin : Plugin {
                 event.plan = it
             }
         }
```
justin-fiedler, line 152: Nit: maybe we should copy the `amplitude.configuration.ingestionMetadata` to prevent unintended modification by enrichment plugins?

core/src/main/java/com/amplitude/core/platform/plugins/ContextPlugin.kt
```diff
@@ -40,6 +40,11 @@ class ContextPlugin : Plugin {
                 event.plan = it
             }
         }
+        event.ingestionMetadata ?: let {
+            amplitude.configuration.ingestionMetadata ?. let {
```
justin-fiedler, line 44: clone here?

justin-fiedler (changes requested):

core/src/main/java/com/amplitude/core/platform/plugins/ContextPlugin.kt
liuyang1520: Thank you @justin-fiedler for reviewing! I create a `clone` method and clone the object here, PTAL whether this is what we want here.

I think timeline enriches the event in sequence, so the clone only prevents concurrent context plugin and enrichment plugin running case? Please let me know if I am wrong.

yuhao900914: Has the same question as Marvin here.
The plugins will be processed in the order of `Before` -> 'Enrichment' -> `Destination`. Feels we cannot prevent the enrichment modification.

yuhao900914 (approved):

core/src/main/java/com/amplitude/core/platform/plugins/ContextPlugin.kt
justin-fiedler: I am concerned about an Enrichment plugin modifying an individual Event's `plan` or `ingestionMetadata` values, which by reference would modify the shared values in `amplitude.configuration` and affect other subsequent events.

justin-fiedler (approved):

core/src/main/java/com/amplitude/core/platform/plugins/ContextPlugin.kt
justin-fiedler: @liuyang1520 let me know if this makes sense. If so you can mark as resolved and merge.

PR #69, opened 2022-09-16
feat: Ingestion metadata update

yuhao900914:
### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/utilities/AndroidStorage.kt
+++ b/android/src/main/java/com/amplitude/android/utilities/AndroidStorage.kt
@@ -8,6 +8,7 @@ import com.amplitude.core.EventCallBack
 import com.amplitude.core.Storage
 import com.amplitude.core.StorageProvider
 import com.amplitude.core.events.BaseEvent
+import com.amplitude.core.events.IngestionMetadata
 import com.amplitude.core.platform.EventPipeline
 import com.amplitude.core.utilities.EventsFileManager
 import com.amplitude.core.utilities.EventsFileStorage
@@ -17,6 +18,7 @@ import com.amplitude.core.utilities.ResponseHandler
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import org.json.JSONArray
+import org.json.JSONObject
 import java.io.File
 
 class AndroidStorage(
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/amplitude/android/utilities/AndroidStorage.kt
```diff
@@ -37,6 +39,12 @@ class AndroidStorage(
 
     override suspend fun writeEvent(event: BaseEvent) {
         eventsFile.storeEvent(JSONUtil.eventToString(event))
+        val amplitudeExtra = event.extra?.get("amplitude")
+        if (amplitudeExtra != null) {
+            try {
+                event.ingestionMetadata = IngestionMetadata.fromJSONObject((amplitudeExtra as Map<String, Any>).get("ingestionMetadata") as JSONObject)
+            } finally { }
```
justin-fiedler, line 46: should this be a `catch`?

justin-fiedler (changes requested):

android/src/main/java/com/amplitude/android/utilities/AndroidStorage.kt
qingzhuozhen: Why add this in this file? This could be at a general place other than here, also I believe the info would be lost if we do it here.

yuhao900914: @justin-fiedler No catch here, because, I think any of the exceptions we should ignore set the ingestionMetadata. Maybe it will be better to have some logged info that says, setting ingestionMetadata error? 

@qingzhuozhen I think this is the place before we store in the storage? Why will this get lost?

yuhao900914: @qingzhuozhen  Which place do you prefer to have the logic? Do you prefer add this logic before writing into the file storage or before sending it to the server?

justin-fiedler: Should this logic be in the AmplitudeDestination plugin rather than storage?

liuyang1520 (approved):

liuyang1520:
The change looks good to me, thanks for keeping trying the solutions! The build [failed](https://github.com/amplitude/Amplitude-Kotlin/actions/runs/3096315411/jobs/5011729113).

justin-fiedler (approved):

core/src/main/java/com/amplitude/core/events/IngestionMetadata.kt
```diff
@@ -44,5 +44,11 @@ open class IngestionMetadata @JvmOverloads constructor(
             val source = jsonObject.optString(AMP_INGESTION_METADATA_SOURCE_VERSION, null)
             return IngestionMetadata(branch, source)
         }
+
+        fun fromMap(map: Map<String, String>): IngestionMetadata {
+            val branch = map.get(AMP_INGESTION_METADATA_SOURCE_NAME) ?: null
```
liuyang1520, line 49: Minor, variable name, `sourceName` and `sourceVersion`

qingzhuozhen (approved):

PR #81, opened 2022-10-03
fix: DXOC-200 lateinit var isBuilt not initialized

bohan-amplitude:
### Summary

change isBuilt to val from lateinit var

### Checklist


liuyang1520 (approved):
LGTM, do we need to update the `isBuilt` in the test too?

qingzhuozhen (approved):

PR #82, opened 2022-10-04
fix: catch filenotfound exception for rollover

bohan-amplitude:
### Summary

catch file not found exceptions for rollover

### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
+++ b/core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
@@ -99,7 +99,13 @@ class EventPipeline(
         uploadChannel.consumeEach {
 
             withContext(amplitude.storageIODispatcher) {
-                storage.rollover()
+                try {
+                    storage.rollover()
+                } catch (e: FileNotFoundException) {
+                    e.message?.let {
+                        amplitude.logger.warn("Event storage file not found: $it")
+                    }
+                }
             }
 
             val eventsData = storage.readEventsContent()
--- a/core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
+++ b/core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
@@ -6,6 +6,7 @@ import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import org.json.JSONArray
 import java.io.File
+import java.io.FileNotFoundException
 import java.io.FileOutputStream
 import java.util.Collections
 import java.util.concurrent.ConcurrentHashMap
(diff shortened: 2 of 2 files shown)


yuhao900914:
Question:
This [issue](https://github.com/amplitude/Amplitude-Kotlin/issues/71) is related to `Too many file opens`.
If we catch the exception here and do nothing. Will we lost those events?

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -165,10 +166,12 @@ class EventsFileManager(
...
-        FileOutputStream(file, true).use {
-            it.write(content)
-            it.flush()
-        }
+        try {
+            FileOutputStream(file, true).use {
+                it.write(content)
+                it.flush()
+            }
+        } catch (e: FileNotFoundException) {}
```
qingzhuozhen, line 174: When would this happen actually? Race condition for writing and deleting? Any thoughts on if we could recover from the issue? 

Another thought is for this kind of exception we could do some debug logging?

bohan-amplitude: I think file not found exceptions are happening when file content was sent out by another coroutine and deleted. This write operation are from rollover() -> finish() -> write a ']' to close the array in the file. Maybe try catch exception from the EventPipeline.update() where called storage.rollover() is better.

qingzhuozhen: OK, I guess as long as we could do a retry sounds good. Maybe better to mimic in a test case.

bohan-amplitude:
@yuhao900914 The file not found exception happen when events already being processed and deleted by another thread. So we don't lose events. Not sure why `too many file opens` the exception seems can't find file by name. I'll clarify with user.

liuyang1520:
Hi team, just a follow up on this fix, @yuhao900914 , @qingzhuozhen , is there any concerns on this fix?

liuyang1520:
Rerun the test, still get the failure, @bohan-amplitude , could you pull the latest code from main? I didn't see the test failures on the recent PRs. Thanks!

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
```diff
@@ -99,7 +99,13 @@ class EventPipeline(
         uploadChannel.consumeEach {
 
             withContext(amplitude.storageIODispatcher) {
-                storage.rollover()
+                try {
+                    storage.rollover()
+                } catch (e: FileNotFoundException) {
+                    e.message?.let {
```
qingzhuozhen, line 105: Do we need to return here? What happens for the later command reading from storage? Would it just recover?

liuyang1520:
Open https://github.com/amplitude/Amplitude-Kotlin/pull/103 and close this one.

PR #85, opened 2022-10-10
fix: clean up permission

qingzhuozhen:
### Summary

Clean up permission

### Checklist

### DIFF
--- a/common-android/src/main/AndroidManifest.xml
+++ b/common-android/src/main/AndroidManifest.xml
@@ -2,5 +2,4 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.amplitude.common.android">
 
-    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
 </manifest>


common-android/src/main/java/com/amplitude/common/android/AndroidContextProvider.kt
```diff
@@ -351,7 +353,17 @@ class AndroidContextProvider(private val context: Context, locationListening: Bo
             if (!isLocationListening) {
                 return null
             }
-            if (!Utils.checkLocationPermissionAllowed(context)) {
+            if (!(
+                ContextCompat.checkSelfPermission(
+                        context,
+                        Manifest.permission.ACCESS_COARSE_LOCATION
```
liuyang1520, line 359: LGTM, so this permission request is moved to the outside app (customer app if they want this track)?

common-android/src/main/AndroidManifest.xml
```diff
@@ -2,5 +2,4 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.amplitude.common.android">
 
-    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
```
yuhao900914, line 5: We only need to include the user-permission in the application, right?

yuhao900914 (approved):

common-android/src/main/AndroidManifest.xml
qingzhuozhen: That's correct.

PR #87, opened 2022-10-23
fix: restore previous session data on app opening

falconandy:
### Summary

1. Current session restore logic didn't work correctly due to Kotlin initialization order (property initializes, `init` blocks, child/parent inheritance). The logic is extracted to a separate `init` block.
2. Added a test for the session restore logic.
3. Session-specific code is moved from coroutines to let it be executed synchronously and avoid possible data-race issues. 
4. Some unnecessary code is removed (including `previousSessionId` property)
5. Fixed bug in `InMemoryStorage` - `contains` was used instead of `containsKey`

### Checklist

### DIFF
--- a/android/src/test/java/com/amplitude/android/AmplitudeTest.kt
+++ b/android/src/test/java/com/amplitude/android/AmplitudeTest.kt
@@ -4,9 +4,13 @@ import android.app.Application
 import android.content.Context
 import com.amplitude.android.plugins.AndroidLifecyclePlugin
 import com.amplitude.common.android.AndroidContextProvider
+import com.amplitude.core.Storage
+import com.amplitude.core.StorageProvider
 import com.amplitude.core.events.BaseEvent
 import com.amplitude.core.platform.EventPlugin
 import com.amplitude.core.platform.Plugin
+import com.amplitude.core.utilities.ConsoleLoggerProvider
+import com.amplitude.core.utilities.InMemoryStorage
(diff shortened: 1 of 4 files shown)


android/src/test/java/com/amplitude/android/AmplitudeTest.kt
```diff
@@ -70,13 +74,14 @@ class AmplitudeTest {
         amplitudeDispatcherField.set(amplitude, dispatcher)
     }
 
-    private fun createConfiguration(minTimeBetweenSessionsMillis: Long? = null): Configuration {
+    private fun createConfiguration(minTimeBetweenSessionsMillis: Long? = null, storageProvider: StorageProvider = InMemoryStorageProvider()): Configuration {
```
qingzhuozhen, line 77: If we keep the test as file storage, anything is failed?

qingzhuozhen (approved):

android/src/test/java/com/amplitude/android/AmplitudeTest.kt
qingzhuozhen: I see. I wonder if we clear the storage file then file storage could work too.

falconandy: Yes, it works - I've checked.

qingzhuozhen: Cool, I guess it is fine now and we could optimize the test case in later changes

PR #89, opened 2022-10-25
fix: move session-specific logic to Timeline

falconandy:
### Summary

All session-related logic is extracted to new Timeline class.

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -4,7 +4,6 @@ import android.content.Context
 import com.amplitude.android.plugins.AndroidContextPlugin
 import com.amplitude.android.plugins.AndroidLifecyclePlugin
 import com.amplitude.core.Amplitude
-import com.amplitude.core.Storage
 import com.amplitude.core.events.BaseEvent
 import com.amplitude.core.platform.plugins.AmplitudeDestination
 import com.amplitude.core.platform.plugins.GetAmpliExtrasPlugin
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -79,59 +79,19 @@ open class Amplitude(
...
-    }
-
     fun onEnterForeground(timestamp: Long) {
-        startNewSessionIfNeeded(timestamp) ?. let {
-            it.forEach { event -> process(event) }
-        }
         inForeground = true
+
+        val dummySessionStartEvent = BaseEvent()
+        dummySessionStartEvent.eventType = START_SESSION_EVENT
```
qingzhuozhen, line 86: So we always trigger this and decide later?

android/src/main/java/com/amplitude/android/Timeline.kt
```diff
@@ -0,0 +1,166 @@
...
+            incomingEvent.timestamp = System.currentTimeMillis()
+        }
+
+        eventMessageChannel.trySend(EventQueueMessage(incomingEvent, (amplitude as Amplitude).inForeground))
+    }
+
+    private suspend fun processEventMessage(message: EventQueueMessage) {
+        val event = message.event
+        var sessionEvents: Iterable<BaseEvent>? = null
+        val eventTimestamp = event.timestamp!!
```
qingzhuozhen, line 45: I believe other event timestamp is bind later in process plugins

android/src/test/java/com/amplitude/android/AmplitudeTest.kt
```diff
@@ -119,6 +119,8 @@ class AmplitudeTest {
             event.eventType = "test event"
             amplitude?.track(event)
             advanceUntilIdle()
+            Thread.sleep(100)
```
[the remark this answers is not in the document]
falconandy: 1. I can increment it to something like `500` but it will increase test duration.
2. I would use more stable ways (something like `???.await()`) but I am not sure what is better/possible to use in Kotlin and do not add notable memory/cpu overhead.

android/src/main/java/com/amplitude/android/Timeline.kt
falconandy: Not sure what you mean. I believe, timestamps should be assigned as soon as possible to have actual live values Now it is here: https://github.com/amplitude/Amplitude-Kotlin/blob/DXOC-297-DiskReadViolation-in-strict-mode/core/src/main/java/com/amplitude/core/Amplitude.kt#L328.
Other code does fallback assignments like:
```
event.timestamp ?: let {
    event.timestamp = System.currentTimeMillis()
}
```

qingzhuozhen: I see, so this is added to cover this. I was worried if this could raiser exceptions.

android/src/test/java/com/amplitude/android/AmplitudeTest.kt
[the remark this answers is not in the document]
qingzhuozhen: I don't believe increasing sleep time is a good option. Not sure if @justin-fiedler or @bgiori know some better ways to mock this in tests in Kotlin though. If not, it maybe fine if you run it several times on local and this test always passes.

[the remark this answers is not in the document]
falconandy: yes, I ran the tests with `Thread.sleep` many times locally and they always pass

android/src/main/java/com/amplitude/android/Timeline.kt
falconandy: `processEventMessage` is invoked only for items from `eventMessageChannel`
https://github.com/amplitude/Amplitude-Kotlin/blob/DXOC-297-DiskReadViolation-in-strict-mode/android/src/main/java/com/amplitude/android/Timeline.kt#L25
the only way to add items to `eventMessageChannel` is `process` function
https://github.com/amplitude/Amplitude-Kotlin/blob/DXOC-297-DiskReadViolation-in-strict-mode/android/src/main/java/com/amplitude/android/Timeline.kt#L34
here `incomingEvent.timestamp = System.currentTimeMillis()` exists to be ensure to have non-null `timestamp`

qingzhuozhen (approved):

PR #95, opened 2022-11-10
fix: add error handling and logging for file load crash issue

liuyang1520:
### Summary
This PR is to address the:
```
Fatal Exception: java.lang.IllegalArgumentException
Malformed \uxxxx encoding
```
image

- I cannot reproduce the same error with setting the `device_id` with a special char, like "时间" (any Chinese char), "\u0000". So I think this should be related to the file path, instead of the content.
- I found an article mentioning the similar problem: https://www.ibm.com/support/pages/error-javalangillegalargumentexception-malformed-uxxxx-encoding-when-start-cognos-analytics

So I think it might be the case that the user is running the app inside the Windows machine within the simulator, and have a customized storage configured with "\user" or "\uxxx" inside the path, then when the file system tries to get that file, it treats the \u as unicode. Note I tried to hack my Android Studio by providing a path containing "\u" when creating the properties file, but it automatically removed all backslashes.

Base on this and the lack of information to reproduce, this PR:
- updates the code to not crash due to this error
- adds loggings information if this case happens so we can understand this better

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -48,7 +48,8 @@ open class Amplitude(
                     instanceName = configuration.instanceName,
                     apiKey = configuration.apiKey,
                     identityStorageProvider = FileIdentityStorageProvider(),
-                    storageDirectory = storageDirectory
+                    storageDirectory = storageDirectory,
+                    logger = configuration.loggerProvider.getLogger(client)
                 )
             )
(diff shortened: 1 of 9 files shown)


qingzhuozhen (commented):
Thanks for adding this, can you also fix the tests?

yuhao900914 (approved):
The logic looks good to me. Please fix the test.

haoliu-amp (approved):

dnshi (approved):

qingzhuozhen (approved):

PR #96, opened 2022-11-19
fix: Add Index out of bound exception handling for PropertiesFile.save()

bohan-amplitude:
### Summary

Expection catch and log for PropertiesFile.save()
Change more general Exception catch for PropertiesFile.load()
Unit test for added function

### Checklist

### DIFF
--- a/id/src/main/java/com/amplitude/id/utilities/PropertiesFile.kt
+++ b/id/src/main/java/com/amplitude/id/utilities/PropertiesFile.kt
@@ -32,8 +32,12 @@ class PropertiesFile(directory: File, key: String, prefix: String, logger: Logge
     }
 
     private fun save() {
-        FileOutputStream(propertiesFile).use {
-            underlyingProperties.store(it, null)
+        try {
+            FileOutputStream(propertiesFile).use {
+                underlyingProperties.store(it, null)
+            }
+        } catch (e: IndexOutOfBoundsException) {
+            logger?.error("Failed to save property file with path ${propertiesFile.absolutePath}, error stacktrace: ${e.stackTraceToString()}")
         }
     }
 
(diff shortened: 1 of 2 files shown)


id/src/main/java/com/amplitude/id/utilities/PropertiesFile.kt
```diff
@@ -32,8 +32,12 @@ class PropertiesFile(directory: File, key: String, prefix: String, logger: Logge
     }
 
     private fun save() {
-        FileOutputStream(propertiesFile).use {
-            underlyingProperties.store(it, null)
+        try {
+            FileOutputStream(propertiesFile).use {
+                underlyingProperties.store(it, null)
+            }
+        } catch (e: IndexOutOfBoundsException) {
```
justin-fiedler, line 39: Is this the only IO error that can happen? We might want to catch all IO errors

bohan-amplitude: Changed to more general exception catch

justin-fiedler (approved):
This should workaround the issue. Were you able to locate the root cause? Is this from lack of file access or is it the Properties have become corrupted.

yuhao900914 (approved):

bohan-amplitude:
> This should workaround the issue. Were you able to locate the root cause? Is this from lack of file access or is it the Properties have become corrupted?

I'll ask if customer can give us some log info that Marvin previous added. It sounds like a random access issue that caused by multithreading

ondrej:
Will the logs you and Marvin added provide more info about the root cause? It looks like we're logging the file path and the stack trace, but the problem appears to be with the values in the properties map.

Do we blanket catch Exception normally in our SDKs? It's normally an anti-pattern since it can potentially hide serious problems that should actually escalate instead of leaving the app in an uncertain state.

PR #107, opened 2023-02-17
feat: add Identify volume reduction

qingzhuozhen:
### Summary

- add identify interceptor feature
- tests pending

### Checklist


core/src/main/java/com/amplitude/core/platform/intercept/IdentifyInterceptFileStorageHandler.kt
```diff
@@ -0,0 +1,118 @@
...
+    override suspend fun fetchAndMergeToNormalEvent(event: BaseEvent): BaseEvent {
+        val eventsData = storage.readEventsContent()
+        if (eventsData.isEmpty()) {
+            return event
+        }
+        val userProperties = mutableMapOf<String, Any?>()
+        for (eventPath in eventsData) {
+            try {
+                val eventsString = storage.getEventsString(eventPath)
+                if (eventsString.isEmpty()) continue
```
justin-fiedler, line 60: nit: single line if's are risky imo. Would prefer braces to prevent unexpected errors based on white space changes.
```suggestion
                if (eventsString.isEmpty()) {
                  continue
                }
```

core/src/main/java/com/amplitude/core/platform/intercept/IdentifyInterceptFileStorageHandler.kt
```diff
@@ -0,0 +1,118 @@
...
+                val events = JSONArray(eventsString).toEvents()
+                val listUserProperties = IdentifyInterceptorUtil.mergeIdentifyList(events)
+                userProperties.putAll(listUserProperties)
+                removeFile(eventPath as String)
+            } catch (e: JSONException) {
+                logger.warn("Identify Merge error: " + e.message)
+                removeFile(eventPath as String)
+            }
```
justin-fiedler, line 94: nit: There is a decent amount of duplicated code between `fetchMergeToIdentifyEvent` and `fetchMergeToNormalEvent`. Could be nice to centralize loading/merging in a single place.

core/src/main/java/com/amplitude/core/platform/intercept/IdentifyInterceptFileStorageHandler.kt
```diff
@@ -0,0 +1,118 @@
...
+import com.amplitude.core.events.IdentifyOperation
+import com.amplitude.core.utilities.FileStorage
+import com.amplitude.core.utilities.toEvents
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import org.json.JSONArray
+import org.json.JSONException
+
+class IdentifyInterceptFileStorageHandler(
```
justin-fiedler, line 14: Do we need dedicated Storage handlers? Could we use an existing event storage class and handle the all the merge logic in IdentifyInterceptor instead?

qingzhuozhen: yeah, that's my first attempt. Break it down mostly on how we handle the events remove. We could discuss later about this.

core/src/main/java/com/amplitude/core/platform/intercept/IdentifyInterceptFileStorageHandler.kt
qingzhuozhen: Good suggestion, created a helper function for it.

android/src/test/java/com/amplitude/android/AmplitudeTest.kt
```diff
@@ -407,7 +408,7 @@ class AmplitudeTest {
 }
 
 class InstanceStorageProvider(private val instance: Storage) : StorageProvider {
-    override fun getStorage(amplitude: com.amplitude.core.Amplitude): Storage {
+    override fun getStorage(amplitude: com.amplitude.core.Amplitude, preifx: String?): Storage {
```
justin-fiedler, line 411: nit: typo

```suggestion
    override fun getStorage(amplitude: com.amplitude.core.Amplitude, prefix: String?): Storage {
```

android/src/test/java/com/amplitude/android/IdentifyInterceptorTest.kt
```diff
@@ -0,0 +1,286 @@
...
+
+    @ExperimentalCoroutinesApi
+    @Before
+    fun setup() {
+        server = MockWebServer()
+        server.start()
+        val context = ApplicationProvider.getApplicationContext<Context>()
+        mockContextProvider()
+
+        val apiKey = "1cc2c1978ebab0f6451112a8f5df4f4e"
```
justin-fiedler, line 48: Is this a real API key? If so we should probably replace with a placeholder, mock.

justin-fiedler (approved):

android/src/test/java/com/amplitude/android/AmplitudeTest.kt
qingzhuozhen: good catch, let me fix it

android/src/test/java/com/amplitude/android/IdentifyInterceptorTest.kt
qingzhuozhen: I guess this is fake, will change to a placeholder

PR #117, opened 2023-04-12
fix: fix log session event when optOut is true

yuhao900914:
### Summary

When set `optOut = true`, we still track the session events. Please check here for the origin [issue](https://amplitude.atlassian.net/browse/DXOC-734?atlOrigin=eyJpIjoiMzBkNGU3NGQ2MjlkNDI4MmE0ZDhmYzhjODFjM2E4Y2QiLCJwIjoiaiJ9). 

### Checklist

* Does your PR have a breaking change?:    No

### DIFF
--- a/android/src/test/java/com/amplitude/android/AmplitudeOtherTest.kt
+++ b/android/src/test/java/com/amplitude/android/AmplitudeOtherTest.kt
@@ -3,13 +3,13 @@ package com.amplitude.android
 import android.content.Context
 import androidx.test.core.app.ApplicationProvider
 import com.amplitude.core.events.BaseEvent
+import io.mockk.spyk
+import io.mockk.verify
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
-import io.mockk.spyk
-import io.mockk.verify
 
 @RunWith(RobolectricTestRunner::class)
 class AmplitudeOtherTest {
--- a/android/src/test/java/com/amplitude/android/AmplitudeOtherTest.kt
+++ b/android/src/test/java/com/amplitude/android/AmplitudeOtherTest.kt
@@ -1,0 +1,51 @@
+package com.amplitude.android
+
+import android.content.Context
+import androidx.test.core.app.ApplicationProvider
+import com.amplitude.core.events.BaseEvent
+import io.mockk.spyk
+import kotlinx.coroutines.ExperimentalCoroutinesApi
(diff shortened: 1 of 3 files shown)


android/src/test/java/com/amplitude/android/AmplitudeOtherTest.kt
```diff
@@ -0,0 +1,51 @@
+package com.amplitude.android
```
justin-fiedler, line 1: Can you please give this a more descriptive name than "AmplitudeOtherTest.kt"? Something like 
 "AmplitudeAndroidTests.kt" or "AmplitudeRobolectricTests.kt" so we can easily add future tests to the correct file.

yuhao900914: Ha, that's a good suggestion. I'm also a little bit struggling with the test name since we already have AmplitudeTest.kt.

yuhao900914: Changed it to `AmplitudeRobolectricTests`.

android/src/test/java/com/amplitude/android/AmplitudeRobolectricTests.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    @ExperimentalCoroutinesApi
+    @Before
+    fun setup() {
+        val context = ApplicationProvider.getApplicationContext<Context>()
+
+        val apiKey = "test-api-key"
+        amplitude = Amplitude(apiKey, context) {
+            this.minTimeBetweenSessionsMillis = 100
+            this.optOut = true
```
justin-fiedler, line 26: nit: to make the test suite more reusable I think we should set `optOut` in the `optOpt` specific test below and only set more common testing config here.

justin-fiedler (approved):

android/src/test/java/com/amplitude/android/AmplitudeRobolectricTests.kt
```diff
@@ -0,0 +1,69 @@
...
+        minTimeBetweenSessionsMillis: Long? = null,
+        optOut: Boolean? = null,
+    ): Configuration {
+
+        return Configuration(
+            apiKey = "api-key",
+            context = context!!,
+            instanceName = "testInstance",
+            loggerProvider = ConsoleLoggerProvider(),
+            identifyInterceptStorageProvider = InMemoryStorageProvider(),
```
qingzhuozhen, line 63: my guess is if you do it with Robolectric you don't have to update this, though I guess it is not a big deal in this case.

qingzhuozhen (approved):

PR #122, opened 2023-05-17
fix: event options should override event properties, refresh last event time on exit foreground

falconandy:
### Summary

1. Event options override event properties
2. `sessionId` is nullable now in event/options.
3. Refresh last event time on exit foreground.
4. Copied session-related tests from Amplitude-Swift.
5. Introduced internal dummy event types (enter foreground, exit foreground) instead of possibly valid session_start/session_end events with null sessionId.

### Checklist


justin-fiedler (approved):
Thanks @falconandy. Tests are great. I like the new dummy foreground background events. Make the logic more clear.

Issue #115, opened 2023-04-05
ConcurrentModificationExceptions


jp2014:
## Expected Behavior

SDK should not cause ConcurrentModificationExceptions

## Current Behavior

SDK Version: 1.7.1

We are seeing a pretty heavy volume of ConcurrentModificationExceptions caused by the Amplitude-Kotlin SDK. We've seen about 5k crashes over the last month related to this.
``` 
Fatal Exception: java.util.ConcurrentModificationException:
       at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:760)
       at java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:793)
       at java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:791)
       at java.util.HashMap.putMapEntries(HashMap.java:511)
       at java.util.HashMap.putAll(HashMap.java:784)
       at com.amplitude.analytics.connector.IdentityStoreImpl$editIdentity$1.updateUserProperties(IdentityStore.kt:75)
       at com.amplitude.android.plugins.AnalyticsConnectorPlugin.execute(AnalyticsConnectorPlugin.kt:45)
       at com.amplitude.core.platform.Mediator.execute(Mediator.java:48)
       at com.amplitude.core.platform.Timeline.applyPlugins(Timeline.kt:37)
       at com.amplitude.core.platform.Timeline.applyPlugins(Timeline.kt:30)
       at com.amplitude.core.platform.Timeline.process(Timeline.kt:16)
       at com.amplitude.android.Timeline.processEventMessage(Timeline.kt:107)
       at com.amplitude.android.Timeline.access$processEventMessage(Timeline.kt:10)
       at com.amplitude.android.Timeline$start$1.invokeSuspend(Timeline.kt:31)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
       at java.lang.Thread.run(Thread.java:920)
```
```
Fatal Exception: java.util.ConcurrentModificationException:
[...]

PR #124, opened 2023-05-31
fix: thread-safe Identify object

falconandy:
### Summary

I tried to reproduce [ConcurrentModificationExceptions bug](https://github.com/amplitude/Amplitude-Kotlin/issues/115). The exception is reproduced if some Identify object is shared and modified between `identify` calls.
Current Identify implementation is not thread-safe.

Now I added synchronized access to mutation methods (`clearAll`, `setUserProperty`) and properties cloning (2 top levels). Not sure if the fix helps in all cases (including [ConcurrentModificationExceptions bug](https://github.com/amplitude/Amplitude-Kotlin/issues/115)).

### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/events/Identify.kt
+++ b/core/src/main/java/com/amplitude/core/events/Identify.kt
@@ -15,10 +15,21 @@ enum class IdentifyOperation(val operationType: String) {
     REMOVE("\$remove")
 }
 
-open class Identify() {
+open class Identify {
 
     private val propertySet: MutableSet<String> = mutableSetOf()
-    val properties = mutableMapOf<String, Any?>()
+
+    private val _properties = mutableMapOf<String, Any?>()
+    val properties: MutableMap<String, Any?>
+        @Synchronized get() {
+            val clonedProperties = _properties.toMutableMap()
+            for ((key, value) in clonedProperties) {
+                if (value is Map<*, *>) {
+                    clonedProperties[key] = value.toMutableMap()
+                }
+            }
+            return clonedProperties
+        }
 
     fun set(property: String, value: Boolean): Identify {
         setUserProperty(IdentifyOperation.SET, property, value)
@@ -540,13 +551,13 @@ open class Identify() {
         return this
     }
 
-    fun clearAll(): Identify {
-        properties.clear()
-        properties.put(IdentifyOperation.CLEAR_ALL.operationType, UNSET_VALUE)
+    @Synchronized fun clearAll(): Identify {
+        _properties.clear()
+        _properties.put(IdentifyOperation.CLEAR_ALL.operationType, UNSET_VALUE)
         return this
     }
 
-    private fun setUserProperty(operation: IdentifyOperation, property: String, value: Any?) {
+    @Synchronized private fun setUserProperty(operation: IdentifyOperation, property: String, value: Any?) {
         if (property.isEmpty()) {
             ConsoleLogger.logger.warn("Attempting to perform operation ${operation.operationType} with a null or empty string property, ignoring")
             return
@@ -556,7 +567,7 @@ open class Identify() {
             return
         }
         // check that clearAll wasn't already used in this Identify
-        if (properties.containsKey(IdentifyOperation.CLEAR_ALL.operationType)) {
+        if (_properties.containsKey(IdentifyOperation.CLEAR_ALL.operationType)) {
             ConsoleLogger.logger.warn("This Identify already contains a \$clearAll operation, ignoring operation %s")
             return
         }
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/amplitude/core/events/Identify.kt
```diff
@@ -15,10 +15,21 @@ enum class IdentifyOperation(val operationType: String) {
     REMOVE("\$remove")
 }
 
-open class Identify() {
+open class Identify {
 
     private val propertySet: MutableSet<String> = mutableSetOf()
-    val properties = mutableMapOf<String, Any?>()
+
+    private val _properties = mutableMapOf<String, Any?>()
```
justin-fiedler, line 22: I think this will fix the first error:
`com.amplitude.analytics.connector.IdentityStoreImpl$editIdentity$1.updateUserProperties(IdentityStore.kt:75)
com.amplitude.android.plugins.AnalyticsConnectorPlugin.execute(AnalyticsConnectorPlugin.kt:45)`

but I still think there still could be errors at:
`com.amplitude.android.plugins.AnalyticsConnectorPlugin.execute(AnalyticsConnectorPlugin.kt:35)`.

Maybe would it be enough to clone the userProperties in `AnalyticsConnectorPlugin.kt` line 30:
` val eventUserProperties: Map<String, Any?>? = event.userProperties?.toMap()`

justin-fiedler: Do we need to make Identify object thread safe? Legacy Android didn't have any synchronization, would be nice to keep our base objects simple too.

I'm not sure if this is limited to Identity but maybe we can instead update the  `IdentityStoreImpl` in Analytics Connector. It already has synchronization code. wdyt?
https://github.com/amplitude/experiment-android-client/blob/main/analytics-connector/src/main/java/com/amplitude/analytics/connector/IdentityStore.kt#L37

falconandy: > but I believe there still could be errors at:
com.amplitude.android.plugins.AnalyticsConnectorPlugin.execute(AnalyticsConnectorPlugin.kt:35).

I was able to reproduce the error on `main` and now it is fixed.
I clone 2 levels (maybe ideally we should clone whole Identify object):
```
val clonedProperties = _properties.toMutableMap() // 1st level
for ((key, value) in clonedProperties) {
    if (value is Map<*, *>) {
        clonedProperties[key] = value.toMutableMap() // 2nd level
    }
}
return clonedProperties
```
1st level cloning fixes  `com.amplitude.android.plugins.AnalyticsConnectorPlugin.execute(AnalyticsConnectorPlugin.kt:35)`
2nd level cloning -  `com.amplitude.analytics.connector.IdentityStoreImpl$editIdentity$1.updateUserProperties(IdentityStore.kt:75)`

> Do we need to make Identify object thread safe? Legacy Android didn't have any synchronization, would be nice to keep these base objects simple too.

In this case we should mention in documentation that an Identify object is not thread-safe and should not be modified after sending to Amplitude.
If developers want to share an Identify object and modify it later they have to implement "cloning" manually.

> Maybe would it be enough to clone the userProperties in AnalyticsConnectorPlugin.kt line 30:
 val eventUserProperties: Map<String, Any?>? = event.userProperties?.toMap()
 I'm not sure if this is limited to Identity but maybe we can instead update the IdentityStoreImpl in Analytics Connector. It already has synchronization code. wdyt?
 
I think, the root issue is [line](https://github.com/amplitude/Amplitude-Kotlin/blob/main/core/src/main/java/com/amplitude/core/Amplitude.kt#L190)
```
event.userProperties = identify.properties
```
Here we set `event.userProperties` to be the same object as `identify.properties`. Then the event is sent asynchronously to another thread/coroutine and `identify.properties` subsequent modifications in main thread can raise the ConcurrentModificationException. For example, in `val eventUserProperties: Map<String, Any?>? = event.userProperties?.toMap()`  ConcurrentModificationException can be raised when the original map `event.userProperties (=== identify.properties)` is enumerated in `toMap()` in the same time when `identify.properties` is modified.

qingzhuozhen: I guess current way could help, though I wonder if leaving the identify as it is now and clone on `event.userProperties = identify.properties` could also work. Ideally the identify object wouldn't have any impact after they send to our method.

falconandy: @qingzhuozhen 
The `identify.properties` is read in a few locations:
```
fun identify(identify: Identify, options: EventOptions? = null): Amplitude {
    ...
    event.userProperties = identify.properties
    ...
}

fun groupIdentify(groupType: String, groupName: String, identify: Identify, options: EventOptions? = null): Amplitude {
    ...
    event.groupProperties = identify.properties
    ...
}

fun setGroup(groupType: String, groupName: String, options: EventOptions? = null): Amplitude {
    ...
    userProperties = identify.properties
    ...
}

fun setGroup(groupType: String, groupName: Array<String>, options: EventOptions? = null): Amplitude {
    ...
    userProperties = identify.properties
    ...
}
```

qingzhuozhen: Gotcha, then it might be easier to do it on get.

justin-fiedler (approved):

PR #125, opened 2023-06-09
fix: do not fetch advertising Id if adid tracking is disabled

falconandy:
Additional note: in Android-Kotlin default value for `limitAdTrackingEnabled` is `true`, in Amplitude-Android - `false`
* https://github.com/amplitude/Amplitude-Kotlin/blob/main/common-android/src/main/java/com/amplitude/common/android/AndroidContextProvider.kt#L51
* https://github.com/amplitude/Amplitude-Android/blob/main/src/main/java/com/amplitude/api/DeviceInfo.java#L56

Maybe if adid tracking is disabled `limitAdTrackingEnabled` should be `null`.

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
+++ b/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
@@ -17,7 +17,11 @@ class AndroidContextPlugin : Plugin {
     override fun setup(amplitude: Amplitude) {
         super.setup(amplitude)
         val configuration = amplitude.configuration as Configuration
-        contextProvider = AndroidContextProvider(configuration.context, configuration.locationListening)
+        contextProvider = AndroidContextProvider(
+            configuration.context,
+            configuration.locationListening,
+            configuration.trackingOptions.shouldTrackAdid()
+        )
         initializeDeviceId(configuration)
     }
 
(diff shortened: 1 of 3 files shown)


justin-fiedler (commented):
Thanks @falconandy. The changes look good to me.

Can you please try fixing the Identify tests some more 🙏  I really want them to be reliable in CI so we can be confident in releases.

justin-fiedler:
> Additional note: in Android-Kotlin default value for limitAdTrackingEnabled is true, in Amplitude-Android - false

Thanks for calling this out. I think this is because ADID usage is no longer recommended, but was at the time of Amplitude-Android's initial release. We probably kept the the legacy SDK as-is for backwards compatibility but changed it in the new SDK along with other breaking changes.

cc @qingzhuozhen

qingzhuozhen (approved):
lg, hope we can fix the build.

Mercy811 (approved):
Want to confirm that to disable tracking ADID, the usage remains the same but ADID is not fetched by the SDK. 

code snippet:
```
val trackingOptions = TrackingOptions()
trackingOptions.disableCity().disableIpAddress().disableLatLng()
amplitude = Amplitude(
    Configuration(
        apiKey = API_KEY,
        context = applicationContext,
        trackingOptions = trackingOptions
    )
)
```

yuhao900914 (approved):

falconandy:
@justin-fiedler 

> Can you please try fixing the Identify tests some more pray I really want them to be reliable in CI so we can be confident in releases.

Fixed (removed global mocking from other tests). Some other tests are still unstable but re-run helps.

@Mercy811 

> Want to confirm that to disable tracking ADID, the usage remains the same but ADID is not fetched by the SDK.

yes, with  `trackingOptions.disableAdid()` the SDK shouldn't fetch ADID

justin-fiedler (approved):

PR #126, opened 2023-06-13
feat: add default events

liuyang1520:
### Summary
This PR mainly does:
- add the default events for the SDK, disabled by default
- update the sample app to have multiple screens

image
image

### Tests
- unit tests
- [manual testing methods](https://amplitude.atlassian.net/wiki/spaces/GROWTH/pages/2109079759/Test+Mobile+SDK+Default+Events)

### Checklist


android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -2,35 +2,78 @@ package com.amplitude.android.plugins
...
+        }
+        application.registerActivityLifecycleCallbacks(this)
     }
 
     override fun onActivityCreated(activity: Activity, bundle: Bundle?) {
+        val configuration = amplitude.configuration as Configuration
+        if (!hasTrackedApplicationLifecycleEvents.getAndSet(true) && configuration.trackingAppLifecycleEvents) {
+            numberOfActivities.set(0)
+            isFirstLaunch.set(true)
+            DefaultEventUtils(amplitude as com.amplitude.android.Amplitude).trackAppUpdatedInstalledEvent(packageInfo)
```
qingzhuozhen, line 42: would this be triggered on each time app open?

android/src/main/java/com/amplitude/android/utilities/DefaultEventUtils.kt
```diff
@@ -0,0 +1,145 @@
...
+                    "[Amplitude] Version" to currentVersion,
+                    "[Amplitude] Build" to currentBuild,
+                ),
+            )
+        } else if (currentBuild != previousBuild) {
+            // Has stored build, but different from current build
+            amplitude.track(
+                EventTypes.APPLICATION_UPDATED,
+                mapOf(
+                    "[Amplitude] Previous Version" to previousVersion,
```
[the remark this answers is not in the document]
liuyang1520: Thanks for reviewing! Do you mean the constants for property types? I can make it for sure.

android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
liuyang1520: `trackAppUpdatedInstalledEvent` will not send event each time app opened:
- for installed event, it will check previous build, if it has (after first time installed), then skip sending installed event
- for updated event, it will compare current build to previous build, if they are the same (no app build upgraded), then skip sending updated event

android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -2,35 +2,78 @@ package com.amplitude.android.plugins
...
     override fun setup(amplitude: Amplitude) {
         super.setup(amplitude)
-        ((amplitude.configuration as Configuration).context as Application).registerActivityLifecycleCallbacks(this)
+
+        val application = (amplitude.configuration as Configuration).context as Application
+        val packageManager: PackageManager = application.packageManager
+        packageInfo = try {
+            packageManager.getPackageInfo(application.packageName, 0)
+        } catch (e: PackageManager.NameNotFoundException) {
+            throw AssertionError("Cannot find package with application.packageName: " + application.packageName)
```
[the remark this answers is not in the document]
liuyang1520: Added in latest commit

android/src/main/java/com/amplitude/android/utilities/DefaultEventUtils.kt
```diff
@@ -0,0 +1,145 @@
...
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+                val intent = activity.intent
+                referrerUri = intent.getParcelableExtra(Intent.EXTRA_REFERRER)
+
+                if (referrerUri == null) {
+                    referrerUri = intent.getStringExtra("android.intent.extra.REFERRER_NAME")?.let {
+                        try {
+                            Uri.parse(it)
+                        } catch (e: ParseException) {
+                            null
```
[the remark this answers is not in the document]
liuyang1520: Added in latest commit

android/src/main/java/com/amplitude/android/utilities/DefaultEventUtils.kt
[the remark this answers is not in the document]
liuyang1520: Update to the consts

android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -2,35 +2,80 @@ package com.amplitude.android.plugins
...
+            isFirstLaunch.set(true)
+            DefaultEventUtils(amplitude as com.amplitude.android.Amplitude).trackAppUpdatedInstalledEvent(packageInfo)
+        }
+        if (configuration.trackingDeepLinks) {
+            DefaultEventUtils(amplitude as com.amplitude.android.Amplitude).trackDeepLinkOpenedEvent(activity)
+        }
     }
 
     override fun onActivityStarted(activity: Activity) {
+        val configuration = amplitude.configuration as Configuration
```
justin-fiedler, line 52: Maybe add a convenience method for getting the typed configuration and amplitude value? Seems to be done in a lot of places.

android/src/test/java/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt
```diff
@@ -0,0 +1,499 @@
...
+        advanceUntilIdle()
+        Thread.sleep(100)
+
+        val tracks = mutableListOf<BaseEvent>()
+        verify(exactly = 0) { mockedPlugin.track(capture(tracks)) }
+        Assertions.assertEquals(0, tracks.count())
+    }
+
+    @Test
+    fun `test screen viewed event is tracked`() = runTest {
```
justin-fiedler, line 316: Nit: to make the tests more concise/maintainable maybe we could use `@TestFactory` to test enabled/disabled with the same logic vs multiple `@Test`s.

https://proandroiddev.com/kotlin-unit-tests-with-parameters-e37aab2b36f6

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -40,6 +40,9 @@ open class Configuration @JvmOverloads constructor(
     var flushEventsOnClose: Boolean = true,
     var minTimeBetweenSessionsMillis: Long = MIN_TIME_BETWEEN_SESSIONS_MILLIS,
     var trackingSessionEvents: Boolean = true,
+    var trackingAppLifecycleEvents: Boolean = false,
+    var trackingDeepLinks: Boolean = false,
+    var trackingScreenViews: Boolean = false,
```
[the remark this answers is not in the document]
justin-fiedler: cc @kevinpagtakhan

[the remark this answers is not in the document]
kevinpagtakhan: yes please! and if we can do the `boolean | Options` patter

android/src/main/java/com/amplitude/android/utilities/DefaultEventUtils.kt
```diff
@@ -0,0 +1,157 @@
...
+import com.amplitude.android.Amplitude
+import com.amplitude.core.Storage
+import kotlinx.coroutines.launch
+
+class DefaultEventUtils(private val amplitude: Amplitude) {
+    object EventTypes {
+        const val APPLICATION_INSTALLED = "[Amplitude] Application Installed"
+        const val APPLICATION_UPDATED = "[Amplitude] Application Updated"
+        const val APPLICATION_OPENED = "[Amplitude] Application Opened"
+        const val APPLICATION_BACKGROUNDED = "[Amplitude] Application Backgrounded"
```
kevinpagtakhan, line 19: Application Closed?

android/src/main/java/com/amplitude/android/Configuration.kt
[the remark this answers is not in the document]
liuyang1520: Thanks @justin-fiedler @kevinpagtakhan! I thought about this, my concerns about the union type are:
- The `trackingSessionEvents` already exists, if we introduce a new `defaultTracking` option with `Boolean` or `Options`, it might introduce some confusion or backward-compatible issues. Plus the session events are enabled by default, while other events are disabled by default.
- Kotlin doesn't support union type by default (https://youtrack.jetbrains.com/issue/KT-13108/Denotable-union-and-intersection-types, currently under discussion whether to support). As far as I have searched, if we want to achieve this, then we would need to expand the type to `Any` (see [example](https://pl.kotl.in/WOeBC8p3C)), or probably introduce some library like [`Either`](https://apidocs.arrow-kt.io/arrow-core/arrow.core/-either/index.html) in Arrow (haven't tried it, not sure whether it is IDE friendly or not).

Let me know your thoughts!

android/src/test/java/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt
liuyang1520: Thanks for sharing this! I checked this article for example too: https://www.arhohuttunen.com/junit-5-kotlin/. I think it might not working in this case, as I want to check the detailed event type/properties when enabled, but not for disabled, and for each event, the flow (trigger) is different. Let me know if I miss any advanced usage of `TestFactory`.

android/src/main/java/com/amplitude/android/utilities/DefaultEventUtils.kt
liuyang1520: Good question, my understanding is application closed event is either not supported or flaky (https://stackoverflow.com/questions/33325035/android-application-kill-event-capture), probably it is why not commonly adopted (see [doc](https://amplitude.atlassian.net/wiki/spaces/GROWTH/pages/2103935062/Default+Event+Tracking+for+Mobile+SDKs#Which-default-events%3F)). Application backgrounded should have similar indication as it relies on `onActivityStopped` hook.

android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
liuyang1520: Updated, thanks!

android/src/main/java/com/amplitude/android/Configuration.kt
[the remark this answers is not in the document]
justin-fiedler: Thanks @liuyang1520. Good points

- "`trackingSessionEvents` already exists" - I think we could mark it as deprecated and give precedence to `defaultTracking` configuration in the case `defaultTracking` is provided by the user.
- I don't think we necessarily need Union types. Instead we could maybe expose some common defaults as static objects to make it easy set a single value. Example (rough) pseudocode

```
open class AttributionOptions @JvmOverloads constructor(
  var excludeReferrers, ...
) {
  companion object {
     const val All = AttributionOptions(...defaultsForAll)
  }
}
class DefaultTrackingOptions @JvmOverloads constructor(
  var attributionOptions: AttributionOptions, ...
) {
  companion object {
     const val All = DefaultTrackingOptions(...defaultsForAll)
     const val None = DefaultTrackingOptions(...defaultsForNone)
  }
}

Amplitude(Configuration(defaultTracking: DefaultTrackingOptions.ALL))
```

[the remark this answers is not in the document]
kevinpagtakhan: i agree with justin on how to move forward with `trackingSessionEvents`. i think the nested config is the "new" way now and would prefer that we have more "new" ways than more "old" ways. sets us up better for 2.x

android/src/test/java/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt
justin-fiedler: I was thinking you could switch on the provided parameter to determine the correct asserts for the results. Seems the initial setup and mocking for both the enabled/disabled test are the same, so would reduce a lot of setup code across the tests.

e.g.

```
(enabled: Bool) ->
// common setup
...
(amplitude.configuration as Configuration).trackingScreenViews = true
...
val tracks = mutableListOf<BaseEvent>()

// different results
if (enabled) { 
  verify { mockedPlugin.track(capture(tracks)) }
          Assertions.assertEquals(1, tracks.count())
          with(tracks[0]) {
              Assertions.assertEquals(DefaultEventUtils.EventTypes.SCREEN_VIEWED, eventType)
              Assertions.assertEquals(eventProperties?.get(DefaultEventUtils.EventProperties.SCREEN_NAME), "test-label")
} else {
    verify(exactly = 0) { mockedPlugin.track(capture(tracks)) }
    Assertions.assertEquals(0, tracks.count())
}
```

android/src/main/java/com/amplitude/android/Configuration.kt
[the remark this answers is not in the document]
liuyang1520: Thanks @justin-fiedler , @kevinpagtakhan! Good idea on the static objects, I will add the deprecate tag to `trackingSessionEvents` and add some common defaults as static objects.

android/src/test/java/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt
liuyang1520: Thanks @justin-fiedler for sharing this! I tried it, seems `dynamicTest` is not compatible with `runTest`, also I feel it actually worse the readability in this case, will keep the current tests then.

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -39,7 +39,9 @@ open class Configuration @JvmOverloads constructor(
     var locationListening: Boolean = true,
     var flushEventsOnClose: Boolean = true,
     var minTimeBetweenSessionsMillis: Long = MIN_TIME_BETWEEN_SESSIONS_MILLIS,
+    @Deprecated("Please use 'defaultTracking' instead.", ReplaceWith("defaultTracking"))
```
justin-fiedler, line 42:
```suggestion
    @Deprecated("Please use 'defaultTracking' instead.", ReplaceWith("defaultTracking.sessions"))
```

android/src/main/java/com/amplitude/android/DefaultTrackingOptions.kt
```diff
@@ -0,0 +1,24 @@
+package com.amplitude.android
+
+class DefaultTrackingOptions(
+    var trackingSessionEvents: Boolean = true,
+    var trackingAppLifecycleEvents: Boolean = false,
+    var trackingDeepLinks: Boolean = false,
+    var trackingScreenViews: Boolean = false
```
justin-fiedler, line 7: I think we should follow the [same naming convent as Browser](https://github.com/amplitude/Amplitude-TypeScript/blob/7d0281ceefd6087b27f854bd3865a49eaf8f745a/packages/analytics-types/src/config/browser.ts#L30-L36)

```suggestion
    var sessions: Boolean = true,
    var appLifecycle: Boolean = false,
    var deepLinks: Boolean = false,
    var screenViews: Boolean = false
```

qingzhuozhen (approved):

android/src/main/java/com/amplitude/android/DefaultTrackingOptions.kt
liuyang1520: @justin-fiedler , good call! I can update, will use `appLifecycles` instead of `appLifecycle` for consistency.

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -39,7 +39,9 @@ open class Configuration @JvmOverloads constructor(
     var locationListening: Boolean = true,
     var flushEventsOnClose: Boolean = true,
     var minTimeBetweenSessionsMillis: Long = MIN_TIME_BETWEEN_SESSIONS_MILLIS,
+    @Deprecated("Please use 'defaultTracking' instead.", ReplaceWith("defaultTracking.sessions"))
```
justin-fiedler, line 42: nit: update comment to match
```suggestion
    @Deprecated("Please use 'defaultTracking.sessions' instead.", ReplaceWith("defaultTracking.sessions"))
```

justin-fiedler (approved):

PR #143, opened 2023-08-18
feat: migrate 'api key' storage data to 'instance name' storage

falconandy:
### Summary

Migrate 'api key' storage data to 'instance name' storage.
Implementation and test are similar to Amplitude-Swift (https://github.com/amplitude/Amplitude-Swift/pull/63)

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -1,13 +1,12 @@
 package com.amplitude.android
 
 import android.content.Context
+import com.amplitude.android.migration.ApiKeyStorageMigration
 import com.amplitude.android.migration.RemnantDataMigration
-import com.amplitude.android.migration.StorageKeyMigration
 import com.amplitude.android.plugins.AnalyticsConnectorIdentityPlugin
 import com.amplitude.android.plugins.AnalyticsConnectorPlugin
 import com.amplitude.android.plugins.AndroidContextPlugin
 import com.amplitude.android.plugins.AndroidLifecyclePlugin
-import com.amplitude.android.utilities.AndroidStorage
 import com.amplitude.core.Amplitude
 import com.amplitude.core.events.BaseEvent
 import com.amplitude.core.platform.plugins.AmplitudeDestination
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/amplitude/core/Storage.kt
```diff
@@ -20,7 +20,9 @@ interface Storage {
 
     suspend fun writeEvent(event: BaseEvent)
 
-    suspend fun write(key: Constants, value: String)
+    fun write(key: Constants, value: String)
```
justin-fiedler, line 23: Why remove suspend? I feel we keep it to make the rest of the interface. wdyt?

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -117,6 +121,22 @@ open class Amplitude(
...
+            StorageKeyMigration(apiKeyStorage, storage, logger).execute()
+        }
+
+        val identifyInterceptStorage = identifyInterceptStorage as? AndroidStorage
+        if (identifyInterceptStorage != null) {
+            val apiKeyStorage = AndroidStorage(configuration.context, configuration.apiKey, logger, identifyInterceptStorage.prefix)
+            StorageKeyMigration(apiKeyStorage, identifyInterceptStorage, logger).execute()
+        }
+    }
+
```
justin-fiedler, line 139: nit: To keep the new logic in Ampliutde.kt to a minium I think it would be nice to extract this logic to a separate class e.g. `ApiKeyStorageMigration(this).execute()` like `RemnantDataMigration`.

justin-fiedler (commented):
Thanks @falconandy. Overall looks good. See comments.

android/src/main/java/com/amplitude/android/Amplitude.kt
falconandy: Extracted

core/src/main/java/com/amplitude/core/Storage.kt
falconandy: @justin-fiedler 
I don't know - maybe https://discuss.kotlinlang.org/t/why-not-make-every-function-suspend-by-default/25224 can help
I've restored `suspend`

justin-fiedler (approved):
Thanks @falconandy. LGTM.

yuhao900914 (approved):

Issue #146, opened 2023-09-05
DataMigration NotImplementedError


brianwernick:
## Expected Behavior
When Amplitude is initialized the `RemnantDataMigration` is skipped when the DB doesn't exist (i.e. no error is thrown)

## Current Behavior
When Amplitude is initialized the `RemnantDataMigration` throws the following exception
```
Fatal Exception: kotlin.NotImplementedError: An operation is not implemented.
       at kotlin.NotImplementedError.<init>(NotImplementedError.java:15)
       at com.amplitude.android.migration.DatabaseStorage.onCreate(DatabaseStorage.kt:53)
       at android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:486)
       at android.database.sqlite.SQLiteOpenHelper.getReadableDatabase(SQLiteOpenHelper.java:415)
       at com.amplitude.android.migration.DatabaseStorage.getValueFromTable(DatabaseStorage.kt:222)
       at com.amplitude.android.migration.DatabaseStorage.getValue(DatabaseStorage.kt:206)
       at com.amplitude.android.migration.RemnantDataMigration.moveDeviceAndUserId(RemnantDataMigration.kt:48)
       at com.amplitude.android.migration.RemnantDataMigration.execute(RemnantDataMigration.kt:36)
       at com.amplitude.android.Amplitude.buildInternal$suspendImpl(Amplitude.kt:55)
       at com.amplitude.android.Amplitude.buildInternal(Amplitude.kt:1)
       at com.amplitude.core.Amplitude$build$built$1.invokeSuspend(Amplitude.kt:107)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:108)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
       at java.lang.Thread.run(Thread.java:919)
```

## Possible Solution
Before attempting to migrate any data the `RemnantDataMigration` should first check if the Database exists and skip the migration if it doesn't (since the `DatabaseStorage` intentionally doesn't support creation).

## Steps to Reproduce
[...]

PR #147, opened 2023-09-06
fix: do not throw NotImplemented error for existing "legacy database"file if the file is not a sqlite file

falconandy:
### Summary

Possible fix for [DataMigration NotImplementedError](https://github.com/amplitude/Amplitude-Kotlin/issues/146).
Data migration already checks if the Database file exists and skips the migration if it doesn't.
But if the file exists but it is not a sqlite database for some reason - NotImplemented error is thrown.
Now migration is skipped in both cases.

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/migration/DatabaseStorage.kt
+++ b/android/src/main/java/com/amplitude/android/migration/DatabaseStorage.kt
@@ -46,11 +46,13 @@ class DatabaseStorage(context: Context, databaseName: String) : SQLiteOpenHelper
     DatabaseConstants.DATABASE_VERSION
 ) {
     private var file: File = context.getDatabasePath(databaseName)
+    private var isValidDatabaseFile = true
     var currentDbVersion: Int = DatabaseConstants.DATABASE_VERSION
         private set
 
     override fun onCreate(db: SQLiteDatabase) {
-        throw NotImplementedError()
+        // File exists but it is not a legacy database for some reason.
+        this.isValidDatabaseFile = false
     }
 
     override fun onUpgrade(db: SQLiteDatabase?, oldVersion: Int, newVersion: Int) {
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/amplitude/android/migration/DatabaseStorage.kt
```diff
@@ -46,11 +46,13 @@ class DatabaseStorage(context: Context, databaseName: String) : SQLiteOpenHelper
     DatabaseConstants.DATABASE_VERSION
 ) {
     private var file: File = context.getDatabasePath(databaseName)
+    private var isValidDatabaseFile = true
     var currentDbVersion: Int = DatabaseConstants.DATABASE_VERSION
         private set
 
     override fun onCreate(db: SQLiteDatabase) {
-        throw NotImplementedError()
+        // File exists but it is not a legacy database for some reason.
```
justin-fiedler, line 54: Curious why this could happen. Could this be a threading issue potentially solved here?
https://github.com/amplitude/Amplitude-Kotlin/pull/144/files

falconandy: I don't know - it is only a hypothesis.

> Could this be a threading issue potentially solved here?

The [PR 144](https://github.com/amplitude/Amplitude-Kotlin/pull/144) only covers file storage, the current bug is related to legacy sqlite storage.

justin-fiedler: Can you please log an error here (by passing in a logger). We should provide a way for the user to know this issue occurred.

falconandy: Added logging.

justin-fiedler (approved):

PR #150, opened 2023-10-13
fix: skip deep link event when url is missing

liuyang1520:
### Summary
fix: skip deep link event when url is missing

This PR filters the deep link event (triggered with activity creation) by only limiting to the case when `intent.data` is available, this can potentially avoid some noisy.

Note this is a breaking change, the existing users will probably see a drop in the event volume, but I think this change aligns more to a bug fix.

### Checklist

### DIFF
--- a/android/src/test/java/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt
+++ b/android/src/test/java/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt
@@ -496,4 +496,30 @@ class AndroidLifecyclePluginTest {
         verify(exactly = 0) { mockedPlugin.track(capture(tracks)) }
         Assertions.assertEquals(0, tracks.count())
     }
+
+    @Test
+    fun `test deep link opened event is not tracked when URL is missing`() = runTest {
+        setDispatcher(testScheduler)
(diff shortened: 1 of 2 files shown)


kelsonpw (approved):

android/src/test/java/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt
```diff
@@ -496,4 +496,30 @@ class AndroidLifecyclePluginTest {
         verify(exactly = 0) { mockedPlugin.track(capture(tracks)) }
         Assertions.assertEquals(0, tracks.count())
     }
+
+    @Test
+    fun `test deep link opened event is not tracked when URL is missing`() = runTest {
+        setDispatcher(testScheduler)
+        configuration.defaultTracking.deepLinks = false
```
Mercy811, line 503: should we enable DET here?

liuyang1520: Yes, good catch! Updated to `true` now. Thanks for reviewing!

Mercy811 (approved):

PR #156, opened 2023-10-31
fix: remove a data race when auto-generated device id is used instead of custom device id

falconandy:
### Summary

1. Adds a new configuration option `deviceId` to prevent device id auto-generation.
2. Adds a new internal `setDeviceId()` method that sets device id if current and expected device ids are equal.
3. Switches `amplitudeDispatcher` to `single thread` mode
4. Adds tests that reproduce the fixed data race.

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
+++ b/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt
@@ -38,19 +38,19 @@ class AndroidContextPlugin : Plugin {
         if (!configuration.newDeviceIdPerInstall && configuration.useAdvertisingIdForDeviceId && !contextProvider.isLimitAdTrackingEnabled()) {
             val advertisingId = contextProvider.advertisingId
             if (advertisingId != null && validDeviceId(advertisingId)) {
-                amplitude.setDeviceId(advertisingId)
+                setDeviceId(advertisingId)
                 return
             }
         }
         if (configuration.useAppSetIdForDeviceId) {
             val appSetId = contextProvider.appSetId
             if (appSetId != null && validDeviceId(appSetId)) {
-                amplitude.setDeviceId("${appSetId}S")
+                setDeviceId("${appSetId}S")
                 return
             }
         }
         val randomId = AndroidContextProvider.generateUUID() + "R"
-        amplitude.setDeviceId(randomId)
+        setDeviceId(randomId)
     }
 
     private fun applyContextData(event: BaseEvent) {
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -46,6 +46,7 @@ open class Configuration @JvmOverloads constructor(
     override var identifyInterceptStorageProvider: StorageProvider = AndroidStorageProvider(),
     override var identityStorageProvider: IdentityStorageProvider = FileIdentityStorageProvider(),
     var migrateLegacyData: Boolean = true,
+    var deviceId: String? = null,
```
justin-fiedler, line 49: If we add deviceId to the configuration we should probably keep it up to date with the current value.

justin-fiedler, line 49: If we add deviceId to the configuration we should probably keep it up to date with the current value.

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -112,6 +112,17 @@ open class Amplitude(
         }
     }
 
+    internal fun setDeviceId(deviceId: String, expectedCurrentDeviceId: String?): Amplitude {
```
justin-fiedler, line 115: Can this logic be done in `core/Amplitude` instead to fix all platforms?

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -112,6 +112,17 @@ open class Amplitude(
...
+        amplitudeScope.launch(amplitudeDispatcher) {
+            isBuilt.await()
+            val currentDeviceId = idContainer.identityManager.getIdentity().deviceId
+            if (currentDeviceId == expectedCurrentDeviceId) {
```
justin-fiedler, line 119: Is the point of this check so that if the `deviceId` has changed (externally) before this method is run, then skip updating it?

I feel like that could have some edge cases. e.g. the deviceId is 1) set to another value then 2) reset to the "expected" value.

Can you think of a way to do this purely by syncronizing/queueing the calls without being reliant on the "expected value"?

justin-fiedler (changes requested):
Thanks @falconandy. See comments.

android/src/main/java/com/amplitude/android/Configuration.kt
falconandy: Removed now. Device sync logic looks pretty complex.

android/src/main/java/com/amplitude/android/Amplitude.kt
falconandy: Removed now. Was added to `android/Amplitude` to be `internal`

android/src/main/java/com/amplitude/android/Amplitude.kt
falconandy: Re-written.

justin-fiedler:
Resolved in #177

PR #160, opened 2023-11-13
fix: correctly format client_upload_time in HTTP request when minIdLength is set

### LINKED ISSUE #159: Malformed Event(s) JSON using 1.12.0 SDK
## Expected Behavior
Generated events JSON should be valid

## Current Behavior
Generated events JSON is invalid. The value of field `client_upload_time` should be a string

```
{
"api_key": "API_KEY",
"client_upload_time": 2023-11-13T12
07: 48.509Z,
"events": [
{
....
}]

```

## Possible Solution
In file `core/src/main/java/com/amplitude/core/utilities/HttpClient.kt#getBodyStr` the else statement (when `minId != null` doesn't add `""` around the field)

## Steps to Reproduce
Log an event to amplitude

## Environment
- Device: Android Emulator sdk_gphone64_arm64
- Device OS and Version: Android OS 13

justin-fiedler:
### Summary

* Fixes #159, [AMP-88899](https://amplitude.atlassian.net/browse/AMP-88899)
* Correctly handles `client_upload_time` when `minIdLength` is set in the configuration
* Added tests for HttpClient

### Checklist

* Does your PR have a breaking change?: No

[AMP-88899]: https://amplitude.atlassian.net/browse/AMP-88899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/test/kotlin/com/amplitude/core/utilities/HttpClientTest.kt
+++ b/core/src/test/kotlin/com/amplitude/core/utilities/HttpClientTest.kt
@@ -46,7 +46,7 @@ class HttpClientTest {
         val event = BaseEvent()
         event.eventType = "test"
 
-        val httpClient = spyk(HttpClient(config));
+        val httpClient = spyk(HttpClient(config))
         every { httpClient.getCurrentTimeMillis() } returns clientUploadTimeMillis
 
         val connection = httpClient.upload()
@@ -59,7 +59,7 @@ class HttpClientTest {
         }
 
         val request: RecordedRequest? = runRequest()
-        val body = request?.body?.readUtf8();
+        val body = request?.body?.readUtf8()
         val result = JSONObject(body)
 
         assertEquals(apiKey, result.getString("api_key"))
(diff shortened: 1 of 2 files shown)


core/src/test/kotlin/com/amplitude/core/utilities/HttpClientTest.kt
```diff
@@ -0,0 +1,108 @@
...
+            serverUrl = server.url("/").toString()
+        )
+        val event = BaseEvent()
+        event.eventType = "test"
+
+        val httpClient = spyk(HttpClient(config))
+        every { httpClient.getCurrentTimeMillis() } returns clientUploadTimeMillis
+
+        val connection = httpClient.upload()
+        System.currentTimeMillis()
```
Mercy811, line 53: It seems that this is not used? And the same for the other test.

Mercy811 (approved):
Thank you Justin for fixing the payload and adding tests.

core/src/test/kotlin/com/amplitude/core/utilities/HttpClientTest.kt
justin-fiedler: Good catch. Thanks

PR #161, opened 2023-11-18
feat: added basic session replay plugin

justin-fiedler:
### Summary
Added basic session replay plugin

* Updated `Plugin` interface to include a `teardown()` method to allow for clean up (similar to Browser SDK)
* Mobile SR plugin
  * Is in it's own module so it can be published separately from the core SDK
  * Adds listener for layout changes
  * Serializes view into a JSON structure based on the current app view hierachy
  * Sends view hierarchy JSON to [MobileSessionReplay server](https://github.com/amplitude/MobileSessionReplayServer/pull/2)
* View component/style support
  * Layouts - positions - supported for most? layouts? 
  * Boundaries - supported for most? components?
  * TextViews - backgroundColor, text, textSize, textColor
  * Buttons - backgroundColor, text, textSize, textColor
  * Toolbar - backgroundColor, text, textSize, textColor

### Checklist


plugin-session-replay-android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.amplitude.android.plugin.sessionreplay">
```
justin-fiedler, line 3: This doesn't really match the module name but I think that `com.amplitude.android.plugin.sessionreplay` is more idiomatic than alternatives.

PR #171, opened 2024-01-02
feat: offline support

Mercy811:
#### Requirements

`ACCESS_NETWORK_STATE` permission is required. If not, this feature will not supported and will fallback to [previous behavior](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/2278555795/Next-gen+SDKs+should+not+drop+events+when+offline#Local-test.1) (causing an error)

#### User interfaces

This feature is enabled by default. 

To opt out of using `AndroidNetworkConnectivityPlugin` and implement custom offline logic, setting `config.offline` to `null` will prevent plugin setup. 
```
// Disable this feature
amplitude = Amplitude(
    Configuration(
        apiKey = AMPLITUDE_API_KEY,
        context = applicationContext,
        offline = AndroidNetworkConnectivityCheckerPlugin.Disabled
    )
)

// Implement your own plugin to toggle offline 
```
#### Implementation

- AndroidNetworkConnectivityPlugin
  - Network connectivity checker will detect network connectivity when setup and set `config.offline` accordingly 
  - Network listener will flush all stored events when back to online
- EventPipeline
  - If no network, it will only save events to storage and will not schedule a flush
  - Otherwise (network connected, no permission, feature disabled), save events and schedule a flush
   

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -1,5 +1,6 @@
 package com.amplitude.android
 
+import AndroidNetworkListener
 import android.content.Context
 import com.amplitude.android.migration.ApiKeyStorageMigration
 import com.amplitude.android.migration.RemnantDataMigration
@@ -7,6 +8,7 @@ import com.amplitude.android.plugins.AnalyticsConnectorIdentityPlugin
 import com.amplitude.android.plugins.AnalyticsConnectorPlugin
 import com.amplitude.android.plugins.AndroidContextPlugin
 import com.amplitude.android.plugins.AndroidLifecyclePlugin
+import com.amplitude.android.utilities.AndroidNetworkConnectivityChecker
 import com.amplitude.core.Amplitude
 import com.amplitude.core.events.BaseEvent
 import com.amplitude.core.platform.plugins.AmplitudeDestination
@@ -16,11 +18,21 @@ import com.amplitude.id.IdentityConfiguration
 import kotlinx.coroutines.launch
 
 open class Amplitude(
-    configuration: Configuration
-) : Amplitude(configuration) {
-
+    configuration: Configuration,
+) : Amplitude(configuration), AndroidNetworkListener.NetworkChangeCallback {
     internal var inForeground = false
     private lateinit var androidContextPlugin: AndroidContextPlugin
+    private var networkListener: AndroidNetworkListener
+    private val networkChangeHandler =
+        object : AndroidNetworkListener.NetworkChangeCallback {
+            override fun onNetworkAvailable() {
+                flush()
+            }
+
+            override fun onNetworkUnavailable() {
+                // Nothing to do so far
+            }
+        }
 
     val sessionId: Long
         get() {
@@ -29,6 +41,9 @@ open class Amplitude(
 
     init {
         registerShutdownHook()
+        networkListener = AndroidNetworkListener((this.configuration as Configuration).context)
+        networkListener.setNetworkChangeCallback(networkChangeHandler)
+        networkListener.startListening()
     }
 
     override fun createTimeline(): Timeline {
@@ -62,7 +77,7 @@ open class Amplitude(
         add(AndroidLifecyclePlugin())
         add(AnalyticsConnectorIdentityPlugin())
         add(AnalyticsConnectorPlugin())
-        add(AmplitudeDestination())
+        add(AmplitudeDestination(AndroidNetworkConnectivityChecker(this.configuration.context, this.logger)))
 
         (timeline as Timeline).start()
     }
@@ -113,11 +128,14 @@ open class Amplitude(
     }
 
     private fun registerShutdownHook() {
-        Runtime.getRuntime().addShutdownHook(object : Thread() {
-            override fun run() {
-                (this@Amplitude.timeline as Timeline).stop()
-            }
-        })
+        Runtime.getRuntime().addShutdownHook(
+            object : Thread() {
+                override fun run() {
+                    (this@Amplitude.timeline as Timeline).stop()
+                    (this@Amplitude.networkListener as AndroidNetworkListener).stopListening()
+                }
+            },
+        )
     }
 
     companion object {
@@ -138,6 +156,14 @@ open class Amplitude(
          * The event type for dummy exit foreground events.
          */
         internal const val DUMMY_EXIT_FOREGROUND_EVENT = "dummy_exit_foreground"
+    }
+
+    override fun onNetworkAvailable() {
+        networkChangeHandler.onNetworkAvailable()
+    }
+
+    override fun onNetworkUnavailable() {
+        networkChangeHandler.onNetworkUnavailable()
     }
 }
 /**
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -19,7 +19,7 @@ import kotlinx.coroutines.launch
 
 open class Amplitude(
     configuration: Configuration,
-) : Amplitude(configuration){
+) : Amplitude(configuration) {
     internal var inForeground = false
     private lateinit var androidContextPlugin: AndroidContextPlugin
     private var networkListener: AndroidNetworkListener
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -1,6 +1,5 @@
 package com.amplitude.android
 
-import AndroidNetworkListener
 import android.content.Context
 import com.amplitude.android.migration.ApiKeyStorageMigration
 import com.amplitude.android.migration.RemnantDataMigration
@@ -23,18 +22,6 @@ open class Amplitude(
 
     internal var inForeground = false
     private lateinit var androidContextPlugin: AndroidContextPlugin
-    private var networkListener: AndroidNetworkListener
-    private val networkChangeHandler =
-        object : AndroidNetworkListener.NetworkChangeCallback {
-            override fun onNetworkAvailable() {
-                configuration.isNetworkConnected = true
-                flush()
-            }
-
-            override fun onNetworkUnavailable() {
-                configuration.isNetworkConnected = false
-            }
-        }
 
     val sessionId: Long
         get() {
@@ -43,9 +30,6 @@ open class Amplitude(
 
     init {
         registerShutdownHook()
-        networkListener = AndroidNetworkListener((this.configuration as Configuration).context)
-        networkListener.setNetworkChangeCallback(networkChangeHandler)
-        networkListener.startListening()
     }
 
     override fun createTimeline(): Timeline {
@@ -134,7 +118,6 @@ open class Amplitude(
         Runtime.getRuntime().addShutdownHook(object : Thread() {
             override fun run() {
                 (this@Amplitude.timeline as Timeline).stop()
-                this@Amplitude.networkListener.stopListening()
             }
         })
     }
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -19,7 +19,7 @@ import kotlinx.coroutines.launch
 
 open class Amplitude(
     configuration: Configuration,
-) : Amplitude(configuration), AndroidNetworkListener.NetworkChangeCallback {
+) : Amplitude(configuration){
     internal var inForeground = false
     private lateinit var androidContextPlugin: AndroidContextPlugin
     private var networkListener: AndroidNetworkListener
@@ -156,14 +156,6 @@ open class Amplitude(
          * The event type for dummy exit foreground events.
          */
         internal const val DUMMY_EXIT_FOREGROUND_EVENT = "dummy_exit_foreground"
-    }
-
-    override fun onNetworkAvailable() {
-        networkChangeHandler.onNetworkAvailable()
-    }
-
-    override fun onNetworkUnavailable() {
-        networkChangeHandler.onNetworkUnavailable()
     }
 }
 /**
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -57,7 +57,9 @@ open class Amplitude(
         }
         this.createIdentityContainer(identityConfiguration)
 
-        add(AndroidNetworkConnectivityCheckerPlugin())
+        if (this.configuration.offline != null) {
+            add(AndroidNetworkConnectivityCheckerPlugin())
+        }
         androidContextPlugin = AndroidContextPlugin()
         add(androidContextPlugin)
         add(GetAmpliExtrasPlugin())
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -18,7 +18,7 @@ import com.amplitude.id.IdentityConfiguration
 import kotlinx.coroutines.launch
 
 open class Amplitude(
-    configuration: Configuration,
+    configuration: Configuration
 ) : Amplitude(configuration) {
     internal var inForeground = false
     private lateinit var androidContextPlugin: AndroidContextPlugin
@@ -128,14 +128,12 @@ open class Amplitude(
     }
 
     private fun registerShutdownHook() {
-        Runtime.getRuntime().addShutdownHook(
-            object : Thread() {
-                override fun run() {
-                    (this@Amplitude.timeline as Timeline).stop()
-                    (this@Amplitude.networkListener as AndroidNetworkListener).stopListening()
-                }
-            },
-        )
+        Runtime.getRuntime().addShutdownHook(object : Thread() {
+            override fun run() {
+                (this@Amplitude.timeline as Timeline).stop()
+                this@Amplitude.networkListener.stopListening()
+            }
+        })
     }
 
     companion object {
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -57,7 +57,7 @@ open class Amplitude(
         }
         this.createIdentityContainer(identityConfiguration)
 
-        if (this.configuration.offline != null) {
+        if (this.configuration.offline != AndroidNetworkConnectivityCheckerPlugin.Disabled) {
             add(AndroidNetworkConnectivityCheckerPlugin())
         }
         androidContextPlugin = AndroidContextPlugin()
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -8,7 +8,7 @@ import com.amplitude.android.plugins.AnalyticsConnectorIdentityPlugin
 import com.amplitude.android.plugins.AnalyticsConnectorPlugin
 import com.amplitude.android.plugins.AndroidContextPlugin
 import com.amplitude.android.plugins.AndroidLifecyclePlugin
-import com.amplitude.android.utilities.AndroidNetworkConnectivityChecker
+import com.amplitude.android.plugins.AndroidNetworkConnectivityCheckerPlugin
 import com.amplitude.core.Amplitude
 import com.amplitude.core.events.BaseEvent
 import com.amplitude.core.platform.plugins.AmplitudeDestination
@@ -20,17 +20,19 @@ import kotlinx.coroutines.launch
 open class Amplitude(
     configuration: Configuration
 ) : Amplitude(configuration) {
+
     internal var inForeground = false
     private lateinit var androidContextPlugin: AndroidContextPlugin
     private var networkListener: AndroidNetworkListener
     private val networkChangeHandler =
         object : AndroidNetworkListener.NetworkChangeCallback {
             override fun onNetworkAvailable() {
+                configuration.isNetworkConnected = true
                 flush()
             }
 
             override fun onNetworkUnavailable() {
-                // Nothing to do so far
+                configuration.isNetworkConnected = false
             }
         }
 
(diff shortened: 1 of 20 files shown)


android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -16,11 +18,21 @@ import com.amplitude.id.IdentityConfiguration
 import kotlinx.coroutines.launch
 
 open class Amplitude(
-    configuration: Configuration
-) : Amplitude(configuration) {
-
+    configuration: Configuration,
+) : Amplitude(configuration), AndroidNetworkListener.NetworkChangeCallback {
```
qingzhuozhen, line 22: why do we need this interface?

qingzhuozhen (changes requested):

android/src/main/java/com/amplitude/android/Amplitude.kt
Mercy811: `NetworkConnectivityChecker` is to check network proactively every time when track an event. `AndroidNetworkListener` is to listen network connection changes and flush when network comes back. If we don't have this interface, stored events will not be flushed automatically when network is back but can only be flushed when another event is tracked.

core/src/main/java/com/amplitude/core/platform/plugins/AmplitudeDestination.kt
```diff
@@ -7,10 +7,11 @@ import com.amplitude.core.events.IdentifyEvent
 import com.amplitude.core.events.RevenueEvent
 import com.amplitude.core.platform.DestinationPlugin
 import com.amplitude.core.platform.EventPipeline
+import com.amplitude.core.platform.NetworkConnectivityChecker
 import com.amplitude.core.platform.intercept.IdentifyInterceptor
 import kotlinx.coroutines.launch
 
-class AmplitudeDestination : DestinationPlugin() {
+class AmplitudeDestination(private val networkConnectivityChecker: NetworkConnectivityChecker? = null) : DestinationPlugin() {
```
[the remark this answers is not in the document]
Mercy811: Could you let me know why you don't like changing the interface here? `AmplitudeDestination` is an internal class so I don't see any cons with it. 

Some reasons of the current design:
- Cleanness. The checker is only needed in `AmplitudeDestination` plugin not `Amplitude` class so it makes more sense to me to pass what the plugin needs to it rather than putting something unnecessary in `Amplitude` class
- Scalability. Android and JVM platform implementations are different. What actually passed in here for Android platform is `AndroidNetworkConnectivityChecker` which implements the interface and uses Android context. If we want to have the same feature for JVM, just implement this interface.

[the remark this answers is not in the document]
qingzhuozhen: It seems causing more friction and not cleanness to me. I am wondering if we could just use a plugin and an extra variable in amplitude instance to decouple and not change much.

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -139,6 +157,14 @@ open class Amplitude(
          */
         internal const val DUMMY_EXIT_FOREGROUND_EVENT = "dummy_exit_foreground"
     }
+
+    override fun onNetworkAvailable() {
```
qingzhuozhen, line 161: @Mercy811 to make the question a bit clear, not quire sure where this method is used and why we need this.

Mercy811: This is not used anywhere but I kept it here only for scalability in case it will be used in the future. In `networkChangeHandler`, `onNetworkAvailable` is left as blank.

core/src/main/java/com/amplitude/core/platform/plugins/AmplitudeDestination.kt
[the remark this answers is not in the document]
Mercy811: I'm afraid that it cannot be implemented by another plugin to prevent the default amplitude destination plugin from scheduling a flush.

android/src/main/java/com/amplitude/android/Amplitude.kt
Mercy811: I see. Remove `NetworkChangeCallback` interface from Amplitude in https://github.com/amplitude/Amplitude-Kotlin/pull/171/commits/70e5ff4ce0766363b78bc1733615cfde9f6a79b6

core/src/main/java/com/amplitude/core/platform/plugins/AmplitudeDestination.kt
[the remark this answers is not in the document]
justin-fiedler: I think I agree with @qingzhuozhen here.

1. Add internal variable `connected` to `android.Amplitude`
2. Create new Before plugin that toggles amplitude.connected
3. We add a new Before plugin to android.Amplitude buildInternal
4. Destination plugin checks amplitude.isConnected

That will keep all the offline detection logic in the Plugin and reduce extra interfaces. wdyt?

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -16,11 +18,21 @@ import com.amplitude.id.IdentityConfiguration
 import kotlinx.coroutines.launch
 
 open class Amplitude(
-    configuration: Configuration
+    configuration: Configuration,
```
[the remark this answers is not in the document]
Mercy811: Looks like we already have [ktlint gradle plugin](https://github.com/amplitude/Amplitude-Kotlin/blob/2df8267d41610438b0b3f2935a9e66a193745d38/build.gradle#L14) specified for the project. `./gradlew ktlintFormat` should format all files.

I'm using [ktlint android studio extension](https://www.paulsblog.dev/ensure-android-code-quality-use-ktlint-in-your-kotlin-project/#using-android-studio-extension). It must be that the rules are different. I reverted lint changes in https://github.com/amplitude/Amplitude-Kotlin/pull/171/commits/acfab11e948e908571ff40c31f4b9d3e79cc562a

android/src/main/java/com/amplitude/android/utilities/AndroidNetworkConnectivityChecker.kt
```diff
@@ -0,0 +1,47 @@
...
+        private const val ACCESS_NETWORK_STATE = "android.permission.ACCESS_NETWORK_STATE"
+    }
+
+    @SuppressLint("MissingPermission")
+    override suspend fun isConnected(): Boolean {
+        // Assume connection and proceed.
+        // Events will be treated like online
+        // regardless network connectivity
+        if (!hasPermission(context, ACCESS_NETWORK_STATE)) {
+            logger.warn("No ACCESS_NETWORK_STATE permission, offline mode is not supported.")
```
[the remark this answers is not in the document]
Mercy811: PR of how to add permission https://github.com/amplitude/amplitude-dev-center/pull/135

Changed in https://github.com/amplitude/Amplitude-Kotlin/pull/171/commits/7e95fd0499fd3e834b0ec86758a632478acaf753

android/src/test/java/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt
```diff
@@ -82,15 +85,19 @@ class AndroidLifecyclePluginTest {
...
+        configuration =
+            Configuration(
+                apiKey = "api-key",
+                context = mockedContext,
+                storageProvider = InMemoryStorageProvider(),
+                loggerProvider = ConsoleLoggerProvider(),
+                identifyInterceptStorageProvider = InMemoryStorageProvider(),
+                identityStorageProvider = IMIdentityStorageProvider(),
+                trackingSessionEvents = false,
+            )
```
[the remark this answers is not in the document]
Mercy811: Reverted. https://github.com/amplitude/Amplitude-Kotlin/pull/171#discussion_r1440734415

android/src/main/java/com/amplitude/android/utilities/AndroidNetworkListener.kt
```diff
@@ -0,0 +1,100 @@
...
+
+class AndroidNetworkListener(private val context: Context) {
+    private var networkCallback: NetworkChangeCallback? = null
+    private var networkCallbackForLowerApiLevels: BroadcastReceiver? = null
+    private var networkCallbackForHigherApiLevels: ConnectivityManager.NetworkCallback? = null
+
+    interface NetworkChangeCallback {
+        fun onNetworkAvailable()
+
+        fun onNetworkUnavailable()
```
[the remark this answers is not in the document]
Mercy811: I'm more into the longer one for explicitness and readability as the interface might be implemented alongside other interfaces.

core/src/main/java/com/amplitude/core/platform/plugins/AmplitudeDestination.kt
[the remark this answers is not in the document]
Mercy811: Thank you @qingzhuozhen and @justin-fiedler. It's more elegant and better decoupled. I implemented network checker as a plugin in https://github.com/amplitude/Amplitude-Kotlin/pull/171/commits/bcd0721544028a5036da43f867ca5a8d096974cf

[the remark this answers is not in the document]
Mercy811: But I make `core.Configuration.isNetworkConnected` `public` as it's used in other modules (`com.amplitude.android.plugins`)

android/src/main/java/com/amplitude/android/Amplitude.kt
Mercy811: After implementing the [plugin architecture](https://github.com/amplitude/Amplitude-Kotlin/pull/171#discussion_r1439992857), this callback is used to change `configuraiton.isNetworkConnected`

android/src/main/java/com/amplitude/android/plugins/AndroidNetworkConnectivityCheckerPlugin.kt
```diff
@@ -0,0 +1,26 @@
...
+    private lateinit var networkConnectivityChecker: AndroidNetworkConnectivityChecker
+
+    override fun setup(amplitude: Amplitude) {
+        super.setup(amplitude)
+        networkConnectivityChecker = AndroidNetworkConnectivityChecker((amplitude.configuration as Configuration).context, amplitude.logger)
+    }
+
+    override fun execute(event: BaseEvent): BaseEvent? {
+        amplitude.amplitudeScope.launch(amplitude.storageIODispatcher) {
+            amplitude.configuration.isNetworkConnected = networkConnectivityChecker.isConnected()
```
qingzhuozhen, line 22: We don't need to do this each time. What if we combine the listener with this plugin, and each time there is a change we just update the variable and trigger the callback(Probably in that way we don't need the callback interface anymore).

core/src/main/java/com/amplitude/core/Configuration.kt
```diff
@@ -30,6 +30,7 @@ open class Configuration @JvmOverloads constructor(
     open var identifyBatchIntervalMillis: Long = IDENTIFY_BATCH_INTERVAL_MILLIS,
     open var identifyInterceptStorageProvider: StorageProvider = InMemoryStorageProvider(),
     open var identityStorageProvider: IdentityStorageProvider = IMIdentityStorageProvider(),
+    var isNetworkConnected: Boolean = true,
```
qingzhuozhen, line 33: If we put the variable on the configuration object, should we call it like this or just offline? I guess when it is set here it will also be allowed to change. A side thing is if we add it in the core configuration to allow custom set then we might also need to add that in the android configuration.

android/src/main/java/com/amplitude/android/plugins/AndroidNetworkConnectivityCheckerPlugin.kt
Mercy811: I think we cannot rely on network change callback considering the case where the APP is started when offline, there is no network change but is still not connected to network. 

IMO Network change callback is used to flush stored events when network is back while network checker is what we should call every time when track.

core/src/main/java/com/amplitude/core/Configuration.kt
Mercy811: I am open to the property name here. @justin-fiedler wdyt?

> I guess when it is set here it will also be allowed to change. A side thing is if we add it in the core configuration to allow custom set then we might also need to add that in the android configuration.

Nice catch. I move it outside of the constructor in https://github.com/amplitude/Amplitude-Kotlin/pull/171/commits/ea4753e9611e9608ba90548a5dbbb1f04511ddd2

Mercy811 (commented):
Move network listener from Amplitude to the network connectivity plugin. So that customers can remove the plugin and have their own offline implementation.

android/src/main/java/com/amplitude/android/utilities/AndroidNetworkConnectivityChecker.kt
```diff
@@ -0,0 +1,49 @@
...
+
+import android.annotation.SuppressLint
+import android.content.Context
+import android.content.pm.PackageManager
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+import android.os.Build
+import com.amplitude.common.Logger
+
+class AndroidNetworkConnectivityChecker(private val context: Context, private val logger: Logger) {
```
justin-fiedler, line 11: Can you please add a test for this class to confirm it works as expected when network state is toggled?

justin-fiedler (changes requested):

android/src/main/java/com/amplitude/android/utilities/AndroidNetworkConnectivityChecker.kt
Mercy811: Added https://github.com/amplitude/Amplitude-Kotlin/blob/0acb081f3dc5b2b1580905ae5309e1e43815e0b3/android/src/test/java/com/amplitude/android/utilities/AndroidNetworkConnectivityCheckerTest.kt

android/src/main/java/com/amplitude/android/plugins/AndroidNetworkConnectivityCheckerPlugin.kt
Mercy811: Added `config.offline` and will check this every time.

Mercy811 (commented):
Check how to remove a plugin.

Plugins are removed by reference. So come up with another way to disable the default plugin by setting `config.offline` to null.

core/src/main/java/com/amplitude/core/platform/Timeline.kt
```diff
@@ -42,6 +42,10 @@ open class Timeline {
     fun remove(plugin: Plugin) {
         // remove all plugins with this name in every category
         plugins.forEach { (_, list) ->
+            val wasRemoved = list.remove(plugin)
```
qingzhuozhen, line 45: why call remove twice?

samples/java-android-app/src/main/java/com/amplitude/android/sample/MainApplication.java
```diff
@@ -84,4 +84,9 @@ public BaseEvent execute(@NonNull BaseEvent event) {
         event.getEventProperties().put("custom android event property", "test");
         return event;
     }
+
+    @Override
```
qingzhuozhen, line 88: @Mercy811 If we don't add this, would this reporting error? Would like to confirm how this interface change would affect users.

android/src/main/java/com/amplitude/android/plugins/AndroidNetworkConnectivityCheckerPlugin.kt
```diff
@@ -0,0 +1,50 @@
...
+                    }
+                }
+            networkListener = AndroidNetworkListener((amplitude.configuration as Configuration).context)
+            networkListener!!.setNetworkChangeCallback(networkChangeHandler)
+            networkListener!!.startListening()
+        }
+    }
+
+    override fun teardown() {
+        if (amplitude.configuration.offline != null) {
```
qingzhuozhen, line 44: seems a workaround to disable other than remove. Remove is a bit more preferred though the hackaround might be ok. Wonder if we still need this check here.

android/src/main/java/com/amplitude/android/plugins/AndroidNetworkConnectivityCheckerPlugin.kt
```diff
@@ -0,0 +1,50 @@
...
+
+class AndroidNetworkConnectivityCheckerPlugin : Plugin {
+    override val type: Plugin.Type = Plugin.Type.Before
+    override lateinit var amplitude: Amplitude
+    internal var networkConnectivityChecker: AndroidNetworkConnectivityChecker? = null
+    internal var networkListener: AndroidNetworkListener? = null
+
+    override fun setup(amplitude: Amplitude) {
+        super.setup(amplitude)
+        if (amplitude.configuration.offline != null) {
```
justin-fiedler, line 18: Is this to allow disabling auto-offline support?

android/src/main/java/com/amplitude/android/plugins/AndroidNetworkConnectivityCheckerPlugin.kt
justin-fiedler: I agree I think I prefer the `amplitude.remove("network-connectivity-detector")`.

I'm guessing most people will leave this as-is.

Also remove Plugin by name would be helpful for other use cases and to match the Browser SDK functionality.

core/src/main/java/com/amplitude/core/platform/Timeline.kt
justin-fiedler: Good catch!

android/src/main/java/com/amplitude/android/plugins/AndroidNetworkConnectivityCheckerPlugin.kt
Mercy811: I'd push back on this.. Remove by name seems to be more like a feature request and is out of the scope of offline feature. Disabling offline by `config.offline = null` also aligns with how we disable default tracking by `config.defaultTracking = DefaultTrackingOptions.NONE`.

core/src/main/java/com/amplitude/core/platform/Timeline.kt
Mercy811: In Kotlin, removing an element from a list that does not exist, nothing dramatic happens and no exception. Nice catch! 

Removed the second `remove()` at https://github.com/amplitude/Amplitude-Kotlin/pull/171/commits/6db25e4c2a92b5e3ce31f920aa950ebb44451073.

samples/java-android-app/src/main/java/com/amplitude/android/sample/MainApplication.java
Mercy811: Yes. This will cause an error 

> Class 'SamplePlugin' must either be declared abstract or implement abstract method 'teardown()' in 'Plugin'.

I did a research and find that it is due to differences in how Java and Kotlin handle default methods in interfaces

- In Kotlin, interfaces can have default implementations for methods, and any Kotlin class implementing such an interface is not required to override these methods with default implementations.
- However, when a Kotlin interface with default method implementations is used in Java, those default implementations are not automatically available. In Java, the class implementing the Kotlin interface is required to provide its own implementation for all methods, including those with default implementations in Kotlin.

So Kotlin app won't be affected, but plugins in Java app needs to add `teardown()`.

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -56,6 +57,9 @@ open class Amplitude(
         }
         this.createIdentityContainer(identityConfiguration)
 
+        if (this.configuration.offline != null) {
```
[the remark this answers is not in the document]
Mercy811: Added a companion variable `Disabled` to `AndroidNetworkConnectivityCheckerPlugin`. Customers can opt out of installing the plugin by 
```
amplitude = Amplitude(
            Configuration(
                apiKey = AMPLITUDE_API_KEY,
                context = applicationContext,
                offline = AndroidNetworkConnectivityCheckerPlugin.Disabled,
            )
        )

```

android/src/test/java/com/amplitude/android/plugins/AndroidNetworkConnectivityCheckerPluginTest.kt
```diff
@@ -0,0 +1,61 @@
...
+    @Test
+    fun `should teardown correctly`() {
+        plugin.setup(amplitude)
+        assertNotNull(plugin.networkListener)
+        plugin.networkListener?.let { networkListener ->
+            spyk(networkListener)
+            plugin.teardown()
+            verify { networkListener.stopListening() }
+        }
+    }
```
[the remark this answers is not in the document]
Mercy811: Unfortunately there is not an easy way to mock toggling network state. But I did a [local test](https://amplitude.atlassian.net/browse/AMP-90600?focusedCommentId=229291).

android/src/main/java/com/amplitude/android/Amplitude.kt
justin-fiedler, line 60: nit: Can you please use the constant in these checks for readability.
```suggestion
        if (this.configuration.offline != AndroidNetworkConnectivityCheckerPlugin.Disabled) {
```

samples/java-android-app/src/main/java/com/amplitude/android/sample/MainApplication.java
justin-fiedler: @Mercy811 can we use default here to avoid interface errors?

Mercy811: Yes! I updated in https://github.com/amplitude/Amplitude-Kotlin/pull/171/commits/f01adb912e3ba6be8a1fc12a7a67f726d4b098f1

justin-fiedler (approved):
Thanks @Mercy811 !

qingzhuozhen (approved):

PR #176, opened 2024-01-26
chore: Add stack traces in exceptions and fix sample app

izaaz:
### Summary

- Logging stack traces when an exception is thrown
- FIxing sample app that was missing an method override

### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/utilities/LoggerExtensions.kt
+++ b/core/src/main/java/com/amplitude/core/utilities/LoggerExtensions.kt
@@ -1,0 +1,12 @@
+package com.amplitude.core.utilities
+
+import com.amplitude.common.Logger
+
+internal fun Exception.logWithStackTrace(logger:Logger, message: String) {
+    this.message?.let {
+        logger.error("$message: $it")
+    }
+    this.stackTrace?.let {
+        logger.error("Stack trace: ${this.stackTraceToString()}")
+    }
+}
--- a/core/src/main/java/com/amplitude/core/utilities/LoggerExtensions.kt
+++ b/core/src/main/java/com/amplitude/core/utilities/LoggerExtensions.kt
@@ -2,7 +2,7 @@ package com.amplitude.core.utilities
 
 import com.amplitude.common.Logger
 
-internal fun Exception.logWithStackTrace(logger:Logger, message: String) {
+internal fun Exception.logWithStackTrace(logger: Logger, message: String) {
     this.message?.let {
         logger.error("$message: $it")
     }
--- a/samples/java-android-app/src/main/java/com/amplitude/android/sample/MainApplication.java
+++ b/samples/java-android-app/src/main/java/com/amplitude/android/sample/MainApplication.java
@@ -84,8 +84,4 @@ class SamplePlugin implements Plugin {
         event.getEventProperties().put("custom android event property", "test");
         return event;
     }
-
-    @Override
-    public void teardown() {
-    }
 }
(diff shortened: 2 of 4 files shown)


samples/java-android-app/src/main/java/com/amplitude/android/sample/MainApplication.java
```diff
@@ -84,4 +84,8 @@ public BaseEvent execute(@NonNull BaseEvent event) {
         event.getEventProperties().put("custom android event property", "test");
         return event;
     }
+
+    @Override
```
justin-fiedler, line 88: @Mercy811 I thought this wouldn't be necessary with the `-Xjvm-default=all` change. Is there something we can do to prevent this being required?

justin-fiedler (approved):

samples/java-android-app/src/main/java/com/amplitude/android/sample/MainApplication.java
Mercy811: it's the replacement for `JvmDefault` not `@Override` unfortunately. https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/#jvmdefaultwithoutcompatibility

Mercy811 (commented):
Thanks Izaaz for adding stack traces. It's very helpful for customers to debug. Could you see if you can remove `teardown()` in the sample project?

samples/java-android-app/src/main/java/com/amplitude/android/sample/MainApplication.java
izaaz: It was throwing a compilation error when I tried running the app locally. Do we add -Xjvm-default=all to our config while running it?

Mercy811: Yes. It's in [build.gradle](https://github.com/amplitude/Amplitude-Kotlin/blob/1ee97e60cd7859755232c3bc7f05693ba6716561/build.gradle#L31). Could you sync gradle?

core/src/main/java/com/amplitude/core/utilities/LoggerExtensions.kt
```diff
@@ -0,0 +1,12 @@
+package com.amplitude.core.utilities
+
+import com.amplitude.common.Logger
+
+internal fun Exception.logWithStackTrace(logger: Logger, message: String) {
+    this.message?.let {
+        logger.error("$message: $it")
+    }
+    this.stackTrace?.let {
+        logger.error("Stack trace: ${this.stackTraceToString()}")
```
qingzhuozhen, line 10: looks great! nit: it might be better to add some Amplitude prefix. Though I believe we could always leave this for later.

qingzhuozhen (approved):

samples/java-android-app/src/main/java/com/amplitude/android/sample/MainApplication.java
izaaz: Thanks for the help @Mercy811. The gradle sync helped remove the error.

core/src/main/java/com/amplitude/core/utilities/LoggerExtensions.kt
izaaz: Yea, I think if we want to do that, we should do that consistently across all our logs. Will leave this as is for now.

Mercy811 (approved):

PR #177, opened 2024-01-29
fix: race condition when setting device id in android

justin-fiedler:
# Summary

## Error
* `amplitude.getDeviceId()` was returning `null` after `amplitude.isBuilt()`.

## Fix
* `deviceId` is now always set before `isBuilt()` is true

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -60,7 +60,12 @@ open class Amplitude(
         if (this.configuration.offline != AndroidNetworkConnectivityCheckerPlugin.Disabled) {
             add(AndroidNetworkConnectivityCheckerPlugin())
         }
-        androidContextPlugin = AndroidContextPlugin()
+        androidContextPlugin = object : AndroidContextPlugin() {
+            override fun setDeviceId(deviceId: String) {
+                // call internal method to set deviceId immediately i.e. dont' wait for build() to complete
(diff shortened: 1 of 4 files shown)


izaaz (approved):

Mercy811 (approved):

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -60,7 +60,12 @@ open class Amplitude(
         if (this.configuration.offline != AndroidNetworkConnectivityCheckerPlugin.Disabled) {
             add(AndroidNetworkConnectivityCheckerPlugin())
         }
-        androidContextPlugin = AndroidContextPlugin()
+        androidContextPlugin = object : AndroidContextPlugin() {
```
justin-fiedler, line 63: Do this for core as well

qingzhuozhen (approved):

android/src/main/java/com/amplitude/android/Amplitude.kt
justin-fiedler: Actually this isn't needed in core since core.ContextPlugin doesn't call setDeviceId()
https://github.com/amplitude/Amplitude-Kotlin/blob/232620cf75e4ece7718678ce6a4dd7553ad859b3/core/src/main/java/com/amplitude/core/platform/plugins/ContextPlugin.kt#L9

qingzhuozhen: Oh, right, core was doing it differently

PR #181, opened 2024-02-22
feat: switch storage for better thread safety

qingzhuozhen:
### Summary

- update storage for better thread safety

### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/Amplitude.kt
+++ b/core/src/main/java/com/amplitude/core/Amplitude.kt
@@ -17,6 +17,7 @@ import com.amplitude.core.platform.plugins.ContextPlugin
 import com.amplitude.core.platform.plugins.GetAmpliExtrasPlugin
 import com.amplitude.core.utilities.AnalyticsEventReceiver
 import com.amplitude.core.utilities.AnalyticsIdentityListener
+import com.amplitude.core.utilities.Diagnostics
 import com.amplitude.eventbridge.EventBridgeContainer
 import com.amplitude.eventbridge.EventChannel
 import com.amplitude.id.IdentityConfiguration
@@ -46,7 +47,7 @@ open class Amplitude internal constructor(
     val amplitudeDispatcher: CoroutineDispatcher = Executors.newCachedThreadPool().asCoroutineDispatcher(),
     val networkIODispatcher: CoroutineDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher(),
     val storageIODispatcher: CoroutineDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher(),
-    val retryDispatcher: CoroutineDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher()
+    val retryDispatcher: CoroutineDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher(),
 ) {
     val timeline: Timeline
     lateinit var storage: Storage
@@ -59,6 +60,7 @@ open class Amplitude internal constructor(
     lateinit var idContainer: IdentityContainer
         private set
     val isBuilt: Deferred<Boolean>
+    val diagnostics = Diagnostics()
 
     init {
         require(configuration.isValid()) { "invalid configuration" }
@@ -82,7 +84,7 @@ open class Amplitude internal constructor(
             instanceName = configuration.instanceName,
             apiKey = configuration.apiKey,
             identityStorageProvider = configuration.identityStorageProvider,
-            logger = logger
+            logger = logger,
         )
     }
 
@@ -98,27 +100,36 @@ open class Amplitude internal constructor(
     protected open fun build(): Deferred<Boolean> {
         val amplitude = this
 
-        val built = amplitudeScope.async(amplitudeDispatcher, CoroutineStart.LAZY) {
-            storage = configuration.storageProvider.getStorage(amplitude)
-            identifyInterceptStorage = configuration.identifyInterceptStorageProvider.getStorage(amplitude, "amplitude-identify-intercept")
-            val identityConfiguration = createIdentityConfiguration()
-            identityStorage = configuration.identityStorageProvider.getIdentityStorage(identityConfiguration)
+        val built =
+            amplitudeScope.async(amplitudeDispatcher, CoroutineStart.LAZY) {
+                storage = configuration.storageProvider.getStorage(amplitude)
+                identifyInterceptStorage =
+                    configuration.identifyInterceptStorageProvider.getStorage(
+                        amplitude,
+                        "amplitude-identify-intercept",
+                    )
+                val identityConfiguration = createIdentityConfiguration()
+                identityStorage = configuration.identityStorageProvider.getIdentityStorage(identityConfiguration)
 
-            amplitude.buildInternal(identityConfiguration)
-            true
-        }
+                amplitude.buildInternal(identityConfiguration)
+                true
+            }
         return built
     }
 
     protected open suspend fun buildInternal(identityConfiguration: IdentityConfiguration) {
         createIdentityContainer(identityConfiguration)
-        EventBridgeContainer.getInstance(configuration.instanceName).eventBridge.setEventReceiver(EventChannel.EVENT, AnalyticsEventReceiver(this))
-        add(object : ContextPlugin() {
-            override fun setDeviceId(deviceId: String) {
-                // set device id immediately, don't wait for isBuilt
-                setDeviceIdInternal(deviceId)
-            }
-        })
+        EventBridgeContainer.getInstance(
+            configuration.instanceName,
+        ).eventBridge.setEventReceiver(EventChannel.EVENT, AnalyticsEventReceiver(this))
+        add(
+            object : ContextPlugin() {
+                override fun setDeviceId(deviceId: String) {
+                    // set device id immediately, don't wait for isBuilt
+                    setDeviceIdInternal(deviceId)
+                }
+            },
+        )
         add(GetAmpliExtrasPlugin())
         add(AmplitudeDestination())
     }
@@ -137,7 +148,11 @@ open class Amplitude internal constructor(
      * @return the Amplitude instance
      */
     @JvmOverloads
-    fun track(event: BaseEvent, options: EventOptions? = null, callback: EventCallBack? = null): Amplitude {
+    fun track(
+        event: BaseEvent,
+        options: EventOptions? = null,
+        callback: EventCallBack? = null,
+    ): Amplitude {
         options ?. let {
             event.mergeEventOptions(it)
         }
@@ -157,7 +172,11 @@ open class Amplitude internal constructor(
      * @return the Amplitude instance
      */
     @JvmOverloads
-    fun track(eventType: String, eventProperties: Map<String, Any?>? = null, options: EventOptions? = null): Amplitude {
+    fun track(
+        eventType: String,
+        eventProperties: Map<String, Any?>? = null,
+        options: EventOptions? = null,
+    ): Amplitude {
         val event = BaseEvent()
         event.eventType = eventType
         event.eventProperties = eventProperties?.toMutableMap()
@@ -177,7 +196,10 @@ open class Amplitude internal constructor(
      * @return the Amplitude instance
      */
     @JvmOverloads
-    fun identify(userProperties: Map<String, Any?>?, options: EventOptions? = null): Amplitude {
+    fun identify(
+        userProperties: Map<String, Any?>?,
+        options: EventOptions? = null,
+    ): Amplitude {
         return identify(convertPropertiesToIdentify(userProperties), options)
     }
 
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -181,23 +245,69 @@ class EventsFileManager(
...
 
     private fun reset() {
         curFile.remove(storageKey)
     }
+
+    private fun handleV1Files() {
+        if (kvs.getLong(storageVersionKey, 1L) > 1L) {
+            return
+        }
+        val unFinishedFiles =
```
izaaz, line 290: Instead of just finishing the files, can we just migrate them over to the new format? That way the read path only needs to worry about the latest format. I'd imagine over time we'd have methods like `handleV1Files`, `handleV2Files` etc and the read path only deals with the latest format

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -122,50 +146,90 @@ class EventsFileManager(
...
+                val isCurrentVersion = content.endsWith("\n")
+                if (isCurrentVersion) {
+                    // handle current version
+                    val events = JSONArray()
+                    content.split("\n").forEach {
+                        if (it.isNotEmpty()) {
+                            try {
+                                events.put(JSONObject(it))
+                            } catch (e: JSONException) {
+                                logger.error("Failed to parse event: $it")
```
izaaz, line 173: It feels unsafe to drop events in the SDK without us being aware of it. Can we still send these events to the server on a different field and have an alert on the server side to alert us when we run into malformed data on the SDK?

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
justin-fiedler: I agree, I would prefer a one time migration to checking the format of each file

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
qingzhuozhen: Let's discuss it later. Is there a field now could be a good fit?

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
qingzhuozhen: I guess it is just we prefer the io at migration time or read time. Originally thought maybe we could save some effort as we need to handle the read io anyway, though agreed it could be more cleanerfor read once we migrate completely.

qingzhuozhen: Migration updated to run at first. Kept the reading for old version files for now just in case anything happened in the migration process.

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
qingzhuozhen: Added the basic diagonostircs, feel free to let me know if we would like to put the interface in a different way. Thanks. cc @justin-fiedler

core/src/main/java/com/amplitude/core/utilities/HttpClient.kt
```diff
@@ -111,13 +111,13 @@ internal class HttpClient(
...
     val inputStream: InputStream?,
-    val outputStream: OutputStream?
+    val outputStream: OutputStream?,
 ) : Closeable {
-
     private lateinit var apiKey: String
     private lateinit var clientUploadTime: String
     private lateinit var events: String
     private var minIdLength: Int? = null
+    private var diagnostics: String? = null
```
izaaz, line 120: Can we make the diagnostics class a typed class? So we know what all diagnostics we are capturing from within the SDK.

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -1,68 +1,93 @@
...
 ) {
-    init {
-        createDirectory(directory)
-    }
-
     private val fileIndexKey = "amplitude.events.file.index.$storageKey"
+    private val storageVersionKey = "amplitude.events.file.version.$storageKey"
+    val filePathSet: MutableSet<String> = Collections.newSetFromMap(ConcurrentHashMap<String, Boolean>())
+    val curFile: MutableMap<String, File> = ConcurrentHashMap<String, File>()
+    private val diagnostics = Diagnostics()
```
izaaz, line 31: Diagnostics feels a little generic. Does it make sense for diagnostics to be scoped to the SDK instead of scoped to the EventsFileManager? I'd imagine we can collect diagnostics from other places too and not just within EventsFileManager. 

If you don't want scope creep and want to just focus on malformed events and just collect that here.

core/src/main/java/com/amplitude/core/utilities/HttpClient.kt
qingzhuozhen: So you mean we pass the diagnostics object here? I guess in HTTP client we could possibly keep the string, while if we could put it in while amplitude client level

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
qingzhuozhen: I am still debating about this and would like to gather the thoughts. Put here for the first iteration and reuse the mutex from the EventsFileManager, though I am considering if I should move this out to make it available to all.

qingzhuozhen: Migrated this on the amplitude instance level so it could possibly cover more if needed.

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -74,9 +99,10 @@ class EventsFileManager(
      */
     fun read(): List<String> {
         // we need to filter out .temp file, since it's operating on the writing thread
-        val fileList = directory.listFiles { _, name ->
-            name.contains(storageKey) && !name.endsWith(".tmp")
-        } ?: emptyArray()
+        val fileList =
+            directory.listFiles { _, name ->
+                name.contains(storageKey) && !name.endsWith(".tmp") && !name.endsWith(".properties")
```
yuhao900914, line 104: what's the file end with `.properties`?

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -320,12 +355,17 @@ open class Amplitude internal constructor(
      * @return the Amplitude instance
      */
     @JvmOverloads
-    fun setGroup(groupType: String, groupName: String, options: EventOptions? = null): Amplitude {
+    fun setGroup(
```
justin-fiedler, line 358: Did the lint settings change? Seems like a lot of style updates

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -122,50 +146,96 @@ class EventsFileManager(
...
+                    val normalizedContent = "[${content.trimStart('[', ',').trimEnd(']', ',')}]"
+                    try {
+                        val jsonArray = JSONArray(normalizedContent)
+                        return@use jsonArray.toString()
+                    } catch (e: JSONException) {
+                        diagnostics.addMalformedEvent(normalizedContent)
+                        logger.error("Failed to parse events: $normalizedContent, dropping file: $filePath")
+                        this.remove(filePath)
+                        return@use ""
+                    }
```
justin-fiedler, line 194: Is this block still needed after the migration?

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -181,23 +251,92 @@ class EventsFileManager(
...
-        file.createNewFile()
-        FileOutputStream(file).use {
-            it.write(content.toByteArray())
-            it.flush()
+    private fun writeEventsToSplitFile(
+        events: List<JSONObject>,
+        file: File,
+        append: Boolean = true,
+    ) {
+        val contents = events.joinToString(separator = "\n", postfix = "\n") { it.toString() }
```
justin-fiedler, line 284: Is there any chance of `\n` in the event name or properties?

core/src/main/java/com/amplitude/core/utilities/Diagnostics.kt
```diff
@@ -0,0 +1,47 @@
+package com.amplitude.core.utilities
+
+import java.util.Collections
+
+class Diagnostics() {
```
justin-fiedler, line 5: Can we make this `internal` for now?

justin-fiedler (commented):
@qingzhuozhen overall looks good. See nits/questions. Thanks for adding diagnostics!

core/src/main/java/com/amplitude/core/Amplitude.kt
qingzhuozhen: Yeah, I wonder if it is for the ktlint version installed is different than before..Would you prefer to revert it back?

core/src/main/java/com/amplitude/core/utilities/Diagnostics.kt
qingzhuozhen: That's what I started first, however, after change it to use in the constructor of EventsFileManager and refer that in android module I need to make this public

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
qingzhuozhen: When it is pure FileStorage, we are using .properties file to store the property. Found this during unit tests so exclude it. In Android this won't happen.

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
qingzhuozhen: In theory it is not needed, I only left this just in case if there is something wrong in migration we still have a backup. Would you prefer we remove this now or it would be better to remove later?

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
qingzhuozhen: Great question! I believe that's possible at least for event or property values. In this write piece this should be fine, though I am a bit more worried on the read part. Will test that a bit more and see how that goes. BTW, any suggestions for handling that?

qingzhuozhen: @justin-fiedler I tried to use the url encode/decode to handle the line break issue and hopefully that helps, let me know your thoughts.

core/src/main/java/com/amplitude/core/Amplitude.kt
justin-fiedler: No preference on the style as long as it stays consistent. Just wanted to check if this was expected.

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
justin-fiedler: Sounds good, I'm ok with this as a safety net for now. Can you maybe add a comment that section is not expected to be called and can be removed later based on diagnostics?

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
justin-fiedler: Thanks. I think the encoding should fix it but am a little concerned about the extra processing for all events. I wonder if there is some other character we can use for the delimiter instead of `\n` that user's are not capable of entering e.g. some special unicode value?

qingzhuozhen: @justin-fiedler I am not sure if there is other values we could use. I guess if we use other value, we still need to handle that value is possibly part of the event or property value. We could discuss later in sync about this.

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
qingzhuozhen: Sure, I can add a comment for this.

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
qingzhuozhen: Just updated to use the new delimiter as discussed.

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
izaaz: Maybe in the else condition we just send the data as malformed events?

qingzhuozhen: @izaaz Just to confirm, would you prefer we return the malformed events or just remove the whole block? In theory this block won't be hit. Though given the edge cases we hit earlier I guess a safety guard won't hurt.

izaaz: I mean we send this data as malformed data to the server so that we can recover if we run into bugs like we did earlier. The problem with safe guards is that we'll start adding it for every new version of the file and this function will soon start getting harder to read.

qingzhuozhen: @izaaz I updated to return the whole content if it falls into malformed case, just kept the handling at first. So we could reduce the case for server to handle. Won't worry much about the new version changes as we could always refine and clean up this part.

justin-fiedler (approved):
Look good. Thanks @qingzhuozhen!

izaaz (approved):

Issue #97, opened 2022-12-07
Event deduplication and insert_id


rajikotta:
there is no way to send insert_id with events in the sdk, is there any other way to prevent duplicate events in amplitude?

kevinpagtakhan:
Hi @rajikotta, thank you for choosing amplitude. The `insert_id` should be populated by default by the SDK with a uuid. Are you seeing otherwise?

rajikotta:
@kevinpagtakhan my original requirement is to avoid duplication over revenue events, can I use insert_id for this purpose

PR #186, opened 2024-03-19
fix: start tracking sessions at init for session replay

justin-fiedler:
### Summary

Sessions are not working as expected with the Session Replay functionality.

The main issue I am trying to solve is ensuring that `amplitude.sessionId` is set after `isBuilt` which was not always the case until now.

Additionally customers have asked for a way to listen to session changes without requiring session events. For this I updated the ObservePlugin to have onSessionIdChanged.

Since Session events can now be fired on instantiation e.g. `val amp = Amplitude(...)` I also added `val amp = Amplitude(Configuration(plugins = listOf(..plugins that run on any events during init))`

Full list of changes:
* No longer migrate session info (sessionId, lastEventId, lastEventTime) - since migration only happens on a new app install, this information can/should be reset.
* Move session logic into its own class
* Remove session logic from timeline
* Process session immediately in constructor, ready before isBuilt
* Add sessionId to state
* Add onSessionIdChanged to ObservePluginMoved session logic to dedicated Session class
* support adding plugins to initial config, this is needed to add properties to session events created during initialization
* centralize time to use SystemTime for easier testing

### Checklist

* Does your PR have a breaking change?:  API - no, Functional - yes - session events now fire immediately rather than on the first event being send.

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -8,6 +8,8 @@ import com.amplitude.android.plugins.AnalyticsConnectorPlugin
 import com.amplitude.android.plugins.AndroidContextPlugin
 import com.amplitude.android.plugins.AndroidLifecyclePlugin
 import com.amplitude.android.plugins.AndroidNetworkConnectivityCheckerPlugin
+import com.amplitude.android.utilities.Session
+import com.amplitude.android.utilities.SystemTime
 import com.amplitude.core.Amplitude
 import com.amplitude.core.events.BaseEvent
 import com.amplitude.core.platform.plugins.AmplitudeDestination
@@ -15,6 +17,7 @@ import com.amplitude.core.platform.plugins.GetAmpliExtrasPlugin
 import com.amplitude.core.utilities.FileStorage
 import com.amplitude.id.IdentityConfiguration
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 
 open class Amplitude(
     configuration: Configuration
@@ -22,10 +25,11 @@ open class Amplitude(
 
     internal var inForeground = false
     private lateinit var androidContextPlugin: AndroidContextPlugin
+    internal val session: Session = Session(configuration)
 
     val sessionId: Long
         get() {
-            return (timeline as Timeline).sessionId
+            return session.sessionId
         }
 
     init {
@@ -33,7 +37,7 @@ open class Amplitude(
     }
 
     override fun createTimeline(): Timeline {
-        return Timeline(configuration.sessionId).also { it.amplitude = this }
+        return Timeline(logger = logger).also { it.amplitude = this }
     }
 
     override fun createIdentityConfiguration(): IdentityConfiguration {
@@ -50,11 +54,19 @@ open class Amplitude(
     }
 
     override suspend fun buildInternal(identityConfiguration: IdentityConfiguration) {
+        session.configure(configuration as Configuration, storage, store, logger)
+        logger.debug("Configured session. Session=$session")
+
         ApiKeyStorageMigration(this).execute()
 
         if ((this.configuration as Configuration).migrateLegacyData) {
             RemnantDataMigration(this).execute()
         }
+
+        // WARNING: Session events need to run after migrations as not to modify `lastEventTime`
+        // Check if we need to start a new session
+        val sessionEvents = session.startNewSessionIfNeeded(SystemTime.getCurrentTimeMillis())
+
         this.createIdentityContainer(identityConfiguration)
 
         if (this.configuration.offline != AndroidNetworkConnectivityCheckerPlugin.Disabled) {
@@ -72,8 +84,21 @@ open class Amplitude(
         add(AnalyticsConnectorIdentityPlugin())
         add(AnalyticsConnectorPlugin())
         add(AmplitudeDestination())
+        val plugins = configuration.plugins
+        if (plugins != null) {
+            for (plugin in plugins) {
+                add(plugin)
+            }
+        }
 
-        (timeline as Timeline).start()
+        val androidTimeline = timeline as Timeline
+        androidTimeline.start(session)
+
+        runBlocking {
+            sessionEvents?.forEach {
+                androidTimeline.processImmediately(it)
+            }
+        }
     }
 
     /**
--- a/android/src/main/java/com/amplitude/android/Timeline.kt
+++ b/android/src/main/java/com/amplitude/android/Timeline.kt
@@ -1,39 +1,25 @@
 package com.amplitude.android
 
+import com.amplitude.android.utilities.Session
+import com.amplitude.android.utilities.SystemTime
+import com.amplitude.common.Logger
 import com.amplitude.core.Storage
 import com.amplitude.core.events.BaseEvent
 import com.amplitude.core.platform.Timeline
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.launch
-import java.util.concurrent.atomic.AtomicLong
 
 class Timeline(
-    private val initialSessionId: Long? = null,
+    private val logger: Logger? = null,
 ) : Timeline() {
     private val eventMessageChannel: Channel<EventQueueMessage> = Channel(Channel.UNLIMITED)
+    private lateinit var session: Session
 
-    private val _sessionId = AtomicLong(initialSessionId ?: -1L)
-    val sessionId: Long
-        get() {
-            return _sessionId.get()
-        }
-
-    internal var lastEventId: Long = 0
-    var lastEventTime: Long = -1L
-
-    internal fun start() {
+    internal fun start(session: Session) {
+        this.session = session
         amplitude.amplitudeScope.launch(amplitude.storageIODispatcher) {
             // Wait until build (including possible legacy data migration) is finished.
             amplitude.isBuilt.await()
-
-            if (initialSessionId == null) {
-                _sessionId.set(
-                    amplitude.storage.read(Storage.Constants.PREVIOUS_SESSION_ID)?.toLongOrNull()
-                        ?: -1
-                )
-            }
-            lastEventId = amplitude.storage.read(Storage.Constants.LAST_EVENT_ID)?.toLongOrNull() ?: 0
-            lastEventTime = amplitude.storage.read(Storage.Constants.LAST_EVENT_TIME)?.toLongOrNull() ?: -1
 
             for (message in eventMessageChannel) {
                 processEventMessage(message)
@@ -47,10 +33,18 @@ class Timeline(
 
     override fun process(incomingEvent: BaseEvent) {
         if (incomingEvent.timestamp == null) {
-            incomingEvent.timestamp = System.currentTimeMillis()
+            incomingEvent.timestamp = SystemTime.getCurrentTimeMillis()
         }
 
         eventMessageChannel.trySend(EventQueueMessage(incomingEvent, (amplitude as Amplitude).inForeground))
+    }
+
+    internal suspend fun processImmediately(incomingEvent: BaseEvent) {
+        if (incomingEvent.timestamp == null) {
+            incomingEvent.timestamp = SystemTime.getCurrentTimeMillis()
+        }
+
+        processEventMessage(EventQueueMessage(incomingEvent, (amplitude as Amplitude).inForeground))
     }
 
     private suspend fun processEventMessage(message: EventQueueMessage) {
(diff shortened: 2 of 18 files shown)


Mercy811 (changes requested):
Thanks Justin for fixing this. 

I can see why `sessionId` is not guaranteed to be set after `isBuilt` as it's part of timeline. To fix this, instead of adding a `session` property in `Amplitude`, how about creating timeline in `build()`? So that we don't need to support adding plugins to initial config which is needed to add properties to session events created during initialization.
```
// core Amplitude
class Amplitude{
    init {
        require(configuration.isValid()) { "invalid configuration" }
        // Stop creating timeline in init
        // timeline = this.createTimeline()
        logger = configuration.loggerProvider.getLogger(this)
        isBuilt = this.build()
        isBuilt.start()
    }

    protected open fun build(): Deferred<Boolean> {
        // create timeline here
        timeline = this.createTimeline()
    }
}
```

android/src/main/java/com/amplitude/android/utilities/Session.kt
```diff
@@ -0,0 +1,118 @@
...
+import com.amplitude.core.State
+import com.amplitude.core.Storage
+import com.amplitude.core.events.BaseEvent
+import java.util.concurrent.atomic.AtomicLong
+
+class Session(
+    private var configuration: Configuration,
+    private var storage: Storage? = null,
+    private var state: State? = null,
+    private var logger: Logger? = null,
```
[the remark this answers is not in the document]
justin-fiedler: Removed

android/src/test/java/com/amplitude/android/AmplitudeSessionTest.kt
```diff
@@ -92,7 +97,7 @@ class AmplitudeSessionTest {
 
         amplitude.isBuilt.await()
 
-        amplitude.track(createEvent(1000, "test event 1"))
+        amplitude.track(createEvent(StartTime, "test event 1"))
         amplitude.track(createEvent(1050, "test event 2"))
```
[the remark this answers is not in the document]
justin-fiedler: Fixed

android/src/test/java/com/amplitude/android/AmplitudeSessionTest.kt
```diff
@@ -110,17 +115,17 @@ class AmplitudeSessionTest {
...
-        Assertions.assertEquals(1000, event.sessionId)
-        Assertions.assertEquals(1000, event.timestamp)
+        Assertions.assertEquals(StartTime, event.sessionId)
+        Assertions.assertEquals(StartTime, event.timestamp)
 
         event = tracks[2]
         Assertions.assertEquals("test event 2", event.eventType)
-        Assertions.assertEquals(1000, event.sessionId)
+        Assertions.assertEquals(StartTime, event.sessionId)
         Assertions.assertEquals(1050, event.timestamp)
```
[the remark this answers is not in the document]
justin-fiedler: Fixed

android/src/test/java/com/amplitude/android/AmplitudeSessionTest.kt
```diff
@@ -604,6 +612,74 @@ class AmplitudeSessionTest {
...
+
+        every { SystemTime.getCurrentTimeMillis() } returns time
+
+        val storageProvider = InstanceStorageProvider(InMemoryStorage())
+        val config = createConfiguration(storageProvider)
+//        config.defaultTracking.sessions = false
+
+        // Create an instance in the background
+        val amplitude1 = Amplitude(config)
+        // amplitude1.logger.logMode = Logger.LogMode.DEBUG
```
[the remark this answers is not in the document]
justin-fiedler: Removed

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -72,8 +84,21 @@ open class Amplitude(
         add(AnalyticsConnectorIdentityPlugin())
         add(AnalyticsConnectorPlugin())
         add(AmplitudeDestination())
+        val plugins = configuration.plugins
```
yuhao900914, line 87: I like the idea of adding plugins through the configurations!
The order of the plugins also matters? should we support ordering the plugins?

android/src/main/java/com/amplitude/android/utilities/Session.kt
```diff
@@ -0,0 +1,99 @@
...
+            sessionEndEvent.eventType = Amplitude.END_SESSION_EVENT
+            sessionEndEvent.timestamp = if (lastEventTime > 0) lastEventTime else null
+            sessionEndEvent.sessionId = sessionId
+            sessionEvents.add(sessionEndEvent)
+        }
+
+        // start new session
+        setSessionId(timestamp)
+        refreshSessionTime(timestamp)
+        if (trackingSessionEvents) {
```
yuhao900914, line 68: How can we fire the identify event before the start session?

android/src/main/java/com/amplitude/android/Amplitude.kt
justin-fiedler: Order is the same as current. Plugins run ordered by type, and then FIFO per type

android/src/main/java/com/amplitude/android/utilities/Session.kt
justin-fiedler: Hmm, not sure yet. I am mostly focused on getting session id working for session replay. Will leave identify fix for another PR.

justin-fiedler:
> I can see why `sessionId` is not guaranteed to be set after `isBuilt` as it's part of timeline. To fix this, instead of adding a `session` property in `Amplitude`, how about creating timeline in `build()`? So that we don't need to support adding plugins to initial config which is needed to add properties to session events created during initialization.

Thanks @Mercy811, I tried moving the session creation into the Timeline.start() but then I need to pass in all the dependencies from Amplitude (configuration, storage, store). I think this implementation is a little cleaner given that. wdyt?

android/src/main/java/com/amplitude/android/utilities/Session.kt
```diff
@@ -0,0 +1,130 @@
...
+        _lastEventTime.set(storage?.read(Storage.Constants.LAST_EVENT_TIME)?.toLongOrNull() ?: DEFAULT_LAST_EVENT_TIME)
+    }
+
+    /**
+     * startNewSessionIfNeeded
+     *
+     * @param timestamp By default this is used as both `sessionId` and `timestamp`
+     * @param sessionId If set, this is used as `sessionId`
+     */
+    suspend fun startNewSessionIfNeeded(timestamp: Long, sessionId: Long? = null): Iterable<BaseEvent>? {
```
izaaz, line 51: How do we handle extending sessions when the app is in foreground?

android/src/main/java/com/amplitude/android/utilities/Session.kt
izaaz: I think at some point we'll need to change the code so that firing of events happens outside session management. 

The logic shoudl be something like 

```
if (session.hasSessionExpired()) {
    fireEndSessionEvent();
    session.createNewSession();
    fireAnyPreSessionInit();
    fireStartSessionEvent();
}
```

izaaz: Not a blocker for this PR

justin-fiedler:
Not sure why the one test is failing. It passes on my local. Debugging it now.

android/src/main/java/com/amplitude/android/utilities/Session.kt
```diff
@@ -0,0 +1,130 @@
...
+    private var state: State? = null,
+) {
+    companion object {
+        const val EMPTY_SESSION_ID = -1L
+        const val DEFAULT_LAST_EVENT_TIME = -1L
+        const val DEFAULT_LAST_EVENT_ID = 0L
+    }
+
+    private val _sessionId = AtomicLong(EMPTY_SESSION_ID)
+    private val _lastEventId = AtomicLong(DEFAULT_LAST_EVENT_ID)
```
izaaz, line 22: Curious - what is the relationship between last event id and a session?

android/src/main/java/com/amplitude/android/utilities/Session.kt
justin-fiedler: Sessions stay active as long as the app is in the foreground, and will set lastEventTime onExitForeground. Some of that logic is done while processing events in the timeline.

android/src/main/java/com/amplitude/android/utilities/Session.kt
justin-fiedler: I felt this made sense to be part of the session since it is tied to the the events sent during the session. wdyt?

android/src/main/java/com/amplitude/android/utilities/Session.kt
izaaz: Ah cool. It's a little different compared to how swift does it. Swift had a flag for inForeground.

android/src/main/java/com/amplitude/android/utilities/Session.kt
izaaz: It does not get reset at the end of a session though right? The reason I brought this up was because another pet peeve I had was that identify suppression also makes our event ids to be skipped. Was hoping to have event id increment happen after identify suppression. 

Not a blocker. We can tackle those problems separately.

izaaz (approved):

android/src/test/java/com/amplitude/android/AmplitudeSessionTest.kt
```diff
@@ -604,6 +632,72 @@ class AmplitudeSessionTest {
...
+        val storageProvider = InstanceStorageProvider(InMemoryStorage())
+        val config = createConfiguration(storageProvider)
+
+        // Create an instance in the background
+        val amplitude1 = Amplitude(config)
+        setDispatcher(amplitude1, testScheduler)
+        amplitude1.isBuilt.await()
+
+        // enter foreground (will start a session)
+        amplitude1.onEnterForeground(time)
```
qingzhuozhen, line 651: one edge case I can think of is when it is initialized first, then create an event before onEnterForeground is triggered. Curious if that's possible or already handled

android/src/main/java/com/amplitude/android/utilities/Session.kt
justin-fiedler: Yeah, I agree, it's a little weird because it is shared across sessions. Originally I was thinking we would reset on a new session.

qingzhuozhen (approved):

Mercy811 (commented):
> I tried moving the session creation into the Timeline.start() but then I need to pass in all the dependencies from Amplitude (configuration, storage, store). I think this implementation is a little cleaner given that. wdyt?

Hi @justin-fiedler, I read the updated code. To not pass all the dependencies  Amplitude (configuration, storage, store), how about accessing them via `amplitude` as core timeline already has amplitude as its class property.

```
// core Timline
open class Timeline {
    lateinit var amplitude: Amplitude // core Amplitude 
}

// android Timeline
class Timeline : Timeline() {
   // should be android Amplitude to enable access `configuration.defaultTracking.sessions` in android Configuration. 
   // Kotlin requires exact type, I did a quick search and it seems to have workaround to use subclass of the original type
   override lateinit var amplitude: Amplitude 
}
```

The biggest concern for me is that right now we have `session` class property in both Amplitude and Timeline and I'd imagine it would be hard to maintain. I feel it's ok to keep `session` in either one of them to have a cleaner logic.

justin-fiedler:
> To not pass all the dependencies Amplitude (configuration, storage, store), how about accessing them via `amplitude` as core timeline already has amplitude as its class property.

Thanks @Mercy811 that made it a lot cleaner.

yuhao900914 (approved):

Mercy811 (approved):
Thank you Justin, this is a huge improvement! 🎉

android/src/main/java/com/amplitude/android/Timeline.kt
```diff
@@ -61,52 +71,42 @@ class Timeline(
         var skipEvent = false
 
         if (event.eventType == Amplitude.START_SESSION_EVENT) {
-            setSessionId(eventSessionId ?: eventTimestamp)
-            refreshSessionTime(eventTimestamp)
+            session.setSessionId(eventSessionId ?: eventTimestamp)
```
yuhao900914, line 74: Question: Does this need? 
When we fire the `START_SESSION_EVENT` we already fire the `setSessionId` which is in android/src/main/java/com/amplitude/android/utilities/Session.kt in line #97?

android/src/test/java/com/amplitude/android/AmplitudeSessionTest.kt
justin-fiedler: Thanks @qingzhuozhen I believe that should be already handled. See the background/foreground tests.

In the case of a init in BG, track event, enter FG, while in BG sessions are subject to `minTimeBetweenSessionsMillis`, so if events are more than that apart a new session will be started.

PR #188, opened 2024-03-25
fix: only take unique error logs and control limit

qingzhuozhen:
### Summary

- only take unique error logs
- control how many error logs we send each time

### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/utilities/Diagnostics.kt
+++ b/core/src/main/java/com/amplitude/core/utilities/Diagnostics.kt
@@ -4,7 +4,7 @@ import java.util.Collections
 
 class Diagnostics() {
     private var malformedEvents: MutableList<String>? = null
-    private var errorLogs: MutableSet<String>? = null
+    private var errorLogs: MutableSet<String> = Collections.synchronizedSet(mutableSetOf())
 
     companion object {
         private const val MAX_ERROR_LOGS = 10
@@ -18,17 +18,14 @@ class Diagnostics() {
     }
 
     fun addErrorLog(log: String) {
-        if (errorLogs == null) {
-            errorLogs = Collections.synchronizedSet(mutableSetOf())
+        errorLogs.add(log)
+        while (errorLogs.size > MAX_ERROR_LOGS) {
+            errorLogs.remove(errorLogs.first())
         }
-        if (errorLogs!!.size >= MAX_ERROR_LOGS) {
-            return
-        }
-        errorLogs?.add(log)
     }
 
     fun hasDiagnostics(): Boolean {
-        return (malformedEvents != null && malformedEvents!!.isNotEmpty()) || (errorLogs != null && errorLogs!!.isNotEmpty())
+        return (malformedEvents != null && malformedEvents!!.isNotEmpty()) || errorLogs.isNotEmpty()
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/amplitude/core/utilities/Diagnostics.kt
```diff
@@ -37,11 +41,13 @@ class Diagnostics() {
             diagnostics["malformed_events"] = malformedEvents!!
         }
         if (errorLogs != null && errorLogs!!.isNotEmpty()) {
-            diagnostics["error_logs"] = errorLogs!!
+            // pick the first MAX_ERROR_LOGS from the set and convert to list
+            val errorLogsToTake = errorLogs!!.take(MAX_ERROR_LOGS)
```
izaaz, line 45: Should we limit the error logs when we add error logs to the set? If the count is greater than max count, there is no need to accumulate more in memory right?

qingzhuozhen: Sure, I can change it to that way as well. I considered this first, then still wants to take error log if unique error logs are more than limit. If we were to worry about the set uniqueness is not enough, I could just limit how many we allowed in our log.

izaaz: Yea I am worried that an error log having a UUID in it will make the uniqueness not sufficient. And in those cases we'll run into a lot of data being saved in memory.

qingzhuozhen: Sounds good, updated.

core/src/main/java/com/amplitude/core/utilities/Diagnostics.kt
```diff
@@ -15,7 +19,10 @@ class Diagnostics() {
 
     fun addErrorLog(log: String) {
         if (errorLogs == null) {
-            errorLogs = Collections.synchronizedList(mutableListOf())
+            errorLogs = Collections.synchronizedSet(mutableSetOf())
+        }
+        if (errorLogs!!.size >= MAX_ERROR_LOGS) {
+            return
```
crleona, line 25: Should we keep the most recent or least recent error logs on overflow? I think this could go either way, but iOS will keep the most recent and purge older entries and we should make the behavior consistent.

core/src/main/java/com/amplitude/core/utilities/Diagnostics.kt
```diff
@@ -4,7 +4,11 @@ import java.util.Collections
 
 class Diagnostics() {
     private var malformedEvents: MutableList<String>? = null
-    private var errorLogs: MutableList<String>? = null
+    private var errorLogs: MutableSet<String>? = null
```
crleona, line 7: Is there any value in maintaining the order of error events?

izaaz, line 7: DOes it make sense to initialize this always and clear the set every time we flush? Helps skip a bunch of null checks and initializations.

izaaz (approved):
Minor comment. Rest LGTM. @crleona's questions are interesting. Its worth making sure swift and kotlin behave similarly at least.

core/src/main/java/com/amplitude/core/utilities/Diagnostics.kt
qingzhuozhen: I guess either way is fine. I can make it to be the same behavior for consistency

core/src/main/java/com/amplitude/core/utilities/Diagnostics.kt
qingzhuozhen: Yeah, could be that way too, was hoping it could save a bit of memory though maybe not that much.

core/src/main/java/com/amplitude/core/utilities/Diagnostics.kt
qingzhuozhen: Updated

crleona (approved):

PR #198, opened 2024-05-30
fix: fix race condition where an event can be processed before a session gets extended

izaaz:
### Summary

This PR fixes a race condition where an event gets processed after the app enters the foreground but before amplitude gets to create a new session.

Instead of setting a flag in Amplitude when the app enters foreground, we now treat all events fired before amplitude's foreground event is fired as background events. 

### Checklist


justin-fiedler (approved):
Thanks @izaaz ! LGTM

PR #199, opened 2024-06-06
feat: add autocapture for element clicks

PouriaAmini:
### Summary
**This PR removes support for Android SDKs 18 and below.**

This PR adds the Autocapture feature to enable automatically tracking element clicks for the legacy Android View system and Jetpack Compose. This functionality is available to both new and existing users but is disabled by default. To enable it, set `userInteractions` in `DefaultTrackingOptions` to `true`:
```
Amplitude(
    Configuration(
        apiKey = AMPLITUDE_API_KEY,
        context = applicationContext,
        defaultTracking = DefaultTrackingOptions(
            userInteractions = true
        )
    )
)
```
or use `DefaultTrackingOptions.ALL`:
```
Amplitude(
    Configuration(
        apiKey = AMPLITUDE_API_KEY,
        context = applicationContext,
        defaultTracking = DefaultTrackingOptions.ALL
    )
)
```
#### New automatically captured event
`[Amplitude] Element Clicked`: Emitted when a user clicks on a clickable element in the legacy Android View system or a tagged clickable `@Composable` functions in Compose.

**Note:** To capture the event in Compose, the clickable `@Composable` functions must include a `Modifier.testTag`, otherwise the event will not be captured. Clickable `@Composable` functions have an `OnClick` lambda or a `Modifier.clickable` modifier (for Jetpack Compose > 1.5).
#### New automatically captured properties
`[Amplitude] Element Class`: (Android Views only) The canonical name of the underlying class of the clickable element, or the simple name as given in the source code if the canonical name is unavailable.

`[Amplitude] Element Resource`: (Android Views only) The entry name for the resource identifier of the clicked element.

`[Amplitude] Element Tag`: (Compose only) The `Modifier.testTag` specified for the clicked @Composable function.

`[Amplitude] Element Source`: (Android Views and Compose) The type of the origin of the clicked element, either "Android Views" for legacy Android Views or "Jetpack Compose" for Compose.

### Checklist

* Does your PR have a breaking change?:  Yes


android/src/main/java/com/amplitude/android/internal/locators/AndroidViewTargetLocator.kt
```diff
@@ -0,0 +1,50 @@
...
+            ?.let { createViewTarget() }
+    }
+
+    private fun View.createViewTarget(): ViewTarget? {
+        try {
+            val className = javaClass.canonicalName ?: javaClass.simpleName ?: null
+            val resourceName = context.resources.getResourceEntryName(id) ?: null
+            return ViewTarget(this, className, resourceName, null, SOURCE)
+        } catch (ignored: Resources.NotFoundException) {
+            return null
```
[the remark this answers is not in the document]
PouriaAmini: Thanks! Added in https://github.com/amplitude/Amplitude-Kotlin/pull/199/commits/7ccff28938eee7f2996bce9e052509ac145fff1d.

android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -55,6 +55,9 @@ class AndroidLifecyclePlugin : Application.ActivityLifecycleCallbacks, Plugin {
         if (androidConfiguration.defaultTracking.screenViews) {
             DefaultEventUtils(androidAmplitude).trackScreenViewedEvent(activity)
         }
+        if (androidConfiguration.defaultTracking.userInteractions) {
+            DefaultEventUtils(androidAmplitude).trackUserInteractionEvent(activity)
```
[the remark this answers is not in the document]
PouriaAmini: I added a `stopUserInteractionEventTracking` to `onActivityPaused`. Also, moved `trackUserInteractionEvent` (now `startUserInteractionEventTracking`) to `onActivityResumed` in https://github.com/amplitude/Amplitude-Kotlin/commit/7ccff28938eee7f2996bce9e052509ac145fff1d.

android/src/main/java/com/amplitude/android/utilities/DefaultEventUtils.kt
```diff
@@ -125,6 +132,19 @@ class DefaultEventUtils(private val amplitude: Amplitude) {
         }
     }
 
+    fun trackUserInteractionEvent(activity: Activity) {
```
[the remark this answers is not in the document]
PouriaAmini: Addressed in https://github.com/amplitude/Amplitude-Kotlin/commit/7ccff28938eee7f2996bce9e052509ac145fff1d as well.

android/build.gradle
```diff
@@ -61,6 +62,8 @@ dependencies {
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.2'
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.2'
     implementation 'com.amplitude:analytics-connector:1.0.0'
+    implementation 'androidx.compose.ui:ui:1.6.7'
```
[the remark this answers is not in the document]
PouriaAmini: I changed that to compileOnly and moved the Compose code to a new `android-compose` module (so no Compose references in the `android` module). I also added a logic to detect if Compose is already enabled. https://github.com/amplitude/Amplitude-Kotlin/pull/199/commits/b0ca2c7aba0045458cb4c6b5ae8c3d3ee6cafc87#diff-ff4c6c993c4216c18d66fab17b786cded7fde9cbd550d21a5322953846d5200eR22-R33

android/build.gradle
```diff
@@ -58,9 +59,11 @@ dependencies {
     embed project(':core')
     embed project(':event-bridge')
     embed project(':id')
+    compileOnly project(':android-compose')
```
justin-fiedler, line 62: I haven't seen `compileOnly` used on a subproject before. Does this selectively include the subproject source code? 

I was expecting you would have the compose libraries here instead (which could be added by the customers app)
```
    compileOnly(compose.runtime)
    compileOnly(compose.ui)
```

PouriaAmini: I just realized that I misunderstood what compileOnly actually does! https://github.com/amplitude/Amplitude-Kotlin/pull/199/commits/ffa1aa1349bd618d7e04cafafeff892fec35fc32 should fix it now.

android/build.gradle
```diff
@@ -61,6 +63,10 @@ dependencies {
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.2'
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.2'
     implementation 'com.amplitude:analytics-connector:1.0.0'
+    implementation "androidx.core:core-ktx:1.7.0"
+    compileOnly compose.runtime
+    compileOnly compose.ui
```
justin-fiedler, line 68: Are they the BOM definitions? Will this only include the indivdual subpackages we need? 

If there is only one or two compose packages required would it be better to include them directly?

PouriaAmini: Here, Compose is handled by the Compose Gradle plugin specified as a classpath. This setup pulls in only the necessary subpackages (like ui, runtime, etc.), so we don't have to specify versions for each one we add now or later.

Since Compose comes from the environment, I'm guessing that setting a specific version might lock the customer to that version. The way it’s set up now, it should work with all Compose versions.

android/src/main/java/com/amplitude/android/DefaultTrackingOptions.kt
```diff
@@ -1,27 +1,34 @@
...
+        // Prebuilt options for easier usage
+        companion object {
+            @JvmField
+            val ALL =
+                DefaultTrackingOptions(
+                    sessions = true,
+                    appLifecycles = true,
+                    deepLinks = true,
+                    screenViews = true,
+                    userInteractions = true,
```
justin-fiedler, line 21: [Can we make this opt-in](https://amplitude.slack.com/archives/C072Y92GTCH/p1718121684448949?thread_ts=1717709659.326509&cid=C072Y92GTCH) until we try with a small group of customers?

justin-fiedler (commented):
Hi @PouriaAmini the PR looks great. 

I think we should do some testing in the wild with a small group of customers before enabling by default in production. Can you please make this feature (temporary) opt-in only?

android/src/main/java/com/amplitude/android/DefaultTrackingOptions.kt
PouriaAmini: I added an annotation to make`userInteractions` require direct opt-in (https://github.com/amplitude/Amplitude-Kotlin/pull/199/commits/5309b73c7dc7c7e40977c51e7480d78eac0d4c60):
```
@OptIn(ExperimentalAmplitudeFeature::class)
    override fun onCreate() {
        super.onCreate()

        amplitude = Amplitude(
            Configuration(
                apiKey = AMPLITUDE_API_KEY,
                context = applicationContext,
                defaultTracking = DefaultTrackingOptions(userInteractions = true)
            )
        )
    }
}
```

Opt-in is only required if `userInteractions` is set in the constructor or directly, not when other default tracking options are enabled/disbaled.

justin-fiedler (approved):
Thanks @PouriaAmini. LGTM. I noticed the build/tests are falling though.

Mercy811 (approved):
Thanks Pouria for adding element clicks autocaputre and using `@ExperimentalAmplitudeFeature`.

PR #200, opened 2024-06-11
ci: Change CI JDK version to 17

PouriaAmini:
### Summary

Changes the JDK version for the build/test workflow.

### Checklist

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -4,6 +4,7 @@ on: [pull_request]
 
 jobs:
   build:
+    name: Build with JDK ${{ matrix.java-version }}
     runs-on: ubuntu-20.04
     strategy:
       matrix:
@@ -33,6 +34,7 @@ jobs:
           retention-days: 7
 
   test-and-lint:
+    name: Test and Lint with JDK 17
     runs-on: ubuntu-20.04
     needs: build
     steps:
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -7,6 +7,17 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v4
+      - name: Set up JDK 11
+        uses: actions/setup-java@v4
+        with:
+          java-version: '11'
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request-test.yml
```diff
@@ -7,10 +7,10 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v4
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
-          java-version: '11'
+          java-version: '17'
```
justin-fiedler, line 13: To see if new changes break with Java 11, what about also running with 11 without the robolectric tests?

PouriaAmini: I added Java 11 back in https://github.com/amplitude/Amplitude-Kotlin/pull/200/commits/467c51068ade551dac725db19100ab6c7d247a4b. This should run the build with JDK 11 as well.

.github/workflows/pull-request-test.yml
```diff
@@ -17,6 +17,17 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'gradle'
+
+      - name: Build
+        run: ./gradlew build
```
justin-fiedler, line 29: This will run the two builds in sequence which is pretty slow. 

Can you please run in parallel instead? The quickest way I think would be to create another workflow that also runs `on: [pull_request]`

PouriaAmini: `strategy.matrix` should run the builds in parallel. Added in https://github.com/amplitude/Amplitude-Kotlin/pull/200/commits/d7f24884a83d603a3f2560cb65fab370fad64bed.

justin-fiedler (approved):
Thanks @PouriaAmini

Mercy811 (approved):
Thanks Pouria for updating it and having both Java 11 and 17 for build to make sure it's compatible with Java 11 as it's still used by half of the applications in 2024.

PR #201, opened 2024-06-12
feat: add scroll and swipe autocapture

PouriaAmini:
### Summary

This PR adds support for scroll and swipe autocapture. The functionality is enabled by default when `userInteractions` is set to `true` in `DefaultTrackingOptions`.
```
@OptIn(ExperimentalAmplitudeFeature::class)
Amplitude(
    Configuration(
        apiKey = AMPLITUDE_API_KEY,
        context = applicationContext,
        defaultTracking = DefaultTrackingOptions(
            userInteractions = true
        )
    )
)
```
#### New automatically captured event
`[Amplitude] Element Scrolled`: Emitted when a user scrolls on a scrollable element in the legacy Android View system or a tagged scrollable `@Composable` functions in Compose.

`[Amplitude] Element Swiped`: Emitted when a user flings on a UI element that was previously scrolled/moved.

Note: To capture the event in Compose, the scrollable `@Composable` functions must include a Modifier.testTag, otherwise the event will not be captured. Scrollable `@Composable` functions have an ScrollBy lambda or a Modifier.scrollable modifier (for Jetpack Compose > 1.5).

#### New automatically captured properties
`[Amplitude] Direction`: (Android Views and Compose) The direction of the scroll or swipe (`Up`, `Right`, `Down`, `Left`).

#### Optimizations around Autocapture implementations
1. Resource name has a fallback now in case a `resourceEntryName` does not exist.
2. Reflection is done lazily for Compose internals.

### Checklist


android/src/main/java/com/amplitude/android/internal/gestures/AutocaptureWindowCallback.kt
```diff
@@ -33,6 +33,13 @@ internal class AutocaptureWindowCallback(
         return super.dispatchTouchEvent(event)
     }
 
+    private fun handleTouchEvent(event: MotionEvent) {
+        gestureDetector.onTouchEvent(event)
+        if (event.actionMasked == MotionEvent.ACTION_UP) {
+            gestureListener.onUp(event)
+        }
```
justin-fiedler, line 40: Can this be handled in `gestureDector.onTouchEvent()` instead?

android/src/main/java/com/amplitude/android/internal/locators/ComposeViewTargetLocator.java
```diff
@@ -81,12 +81,19 @@ public ViewTarget locate(
...
-                                    lastKnownTag = (String) entry.getValue();
-                                }
+                            switch (key) {
+                                case "ScrollBy":
+                                    isScrollable = true;
+                                    break;
+                                case "OnClick":
+                                    isClickable = true;
+                                    break;
+                                case "SentryTag":
```
justin-fiedler, line 91:
```suggestion
```

android/src/main/java/com/amplitude/android/internal/gestures/AutocaptureWindowCallback.kt
PouriaAmini: Since GestureDetector is a built-in class and gestureDector is an instance of that, I don't think there is a way to add this logic to `gestureDector.onTouchEvent()` instead.

android/src/main/java/com/amplitude/android/internal/locators/ComposeViewTargetLocator.java
PouriaAmini: Nice catch! Fixed in https://github.com/amplitude/Amplitude-Kotlin/pull/201/commits/c303099245c55d870cae86d3e81136e4dd7d6cc3.

android/src/main/java/com/amplitude/android/internal/gestures/AutocaptureWindowCallback.kt
justin-fiedler: Ah sorry I misread that these were both `gestureDetector` vs `gestureListener`

android/src/main/java/com/amplitude/android/internal/gestures/AutocaptureGestureListener.kt
```diff
@@ -62,17 +64,133 @@ class AutocaptureGestureListener(
...
+    }
+
+    private fun capture(
+        target: ViewTarget,
+        eventType: String,
+        additionalProperties: Map<String, Any?> = emptyMap(),
+    ) {
+        mapOf(
+            ELEMENT_CLASS to target.className,
+            ELEMENT_RESOURCE to target.recourseName,
```
justin-fiedler, line 143: nit:
```suggestion
            ELEMENT_RESOURCE to target.resourceName,
```

android/src/test/java/com/amplitude/android/internal/gestures/AutocaptureGestureListenerClickTest.kt
```diff
@@ -145,12 +142,12 @@ class AutocaptureGestureListenerClickTest {
 
         verify(exactly = 1) {
             fixture.track(
-                ELEMENT_CLICKED,
+                "[Amplitude] Element Clicked",
```
justin-fiedler, line 145: Why the switch to hardcorded strings in tests?

justin-fiedler (approved):
Thanks @PouriaAmini. LGTM. See nits.

android/src/test/java/com/amplitude/android/internal/gestures/AutocaptureGestureListenerClickTest.kt
PouriaAmini: This is just to make sure that we don't accidentally change the event/property name.

android/src/main/java/com/amplitude/android/internal/gestures/AutocaptureGestureListener.kt
PouriaAmini: 😅
Fixed in https://github.com/amplitude/Amplitude-Kotlin/pull/201/commits/c10fc5dc8d3444c64f4534eae7e639f0d9d9dee3.

PR #207, opened 2024-07-29
feat: add autocapture in configs

PouriaAmini:
### Summary

This PR deprecates the `defaultTracking` option in the SDK configurations and adds `autocapture`.

### Checklist


android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -41,18 +41,22 @@ class AndroidLifecyclePlugin : Application.ActivityLifecycleCallbacks, Plugin {
     }
 
     override fun onActivityCreated(activity: Activity, bundle: Bundle?) {
-        if (!hasTrackedApplicationLifecycleEvents.getAndSet(true) && androidConfiguration.defaultTracking.appLifecycles) {
+        @Suppress("DEPRECATION")
+        val trackingAppLifecycle = androidConfiguration.defaultTracking.appLifecycles || androidConfiguration.autocapture.appLifecycles
```
yuhao900914, line 45: Maybe we can have a function to merge the default tracking and autocapture config together, so all the if statements can be shorter.

PouriaAmini: Thanks for the callout! I've tried doing that before, but since it's possible to change defaultTracking options later after the app has initialized Amplitude, merging default tracking and autocapture might discard those changes unintentionally, which would cause a silent bug. For example, if the app does:
```
configs.defaultTracking.sessions = true
```
after initializing Amplitude, there doesn't seem to be a straightforward way to capture this change and update autocapture accordingly.

crleona: Use one canonical representation of the options internally, probably the auto capture variant. Everything else should bridge to that.

PouriaAmini: 124663d0c358e0768c183e18530def4f75efd351 should fix this.

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -39,9 +39,10 @@ open class Configuration @JvmOverloads constructor(
     var locationListening: Boolean = true,
     var flushEventsOnClose: Boolean = true,
     var minTimeBetweenSessionsMillis: Long = MIN_TIME_BETWEEN_SESSIONS_MILLIS,
-    @Deprecated("Please use 'defaultTracking.sessions' instead.", ReplaceWith("defaultTracking.sessions"))
-    var trackingSessionEvents: Boolean = true,
-    var defaultTracking: DefaultTrackingOptions = DefaultTrackingOptions(),
+    trackingSessionEvents: Boolean = true,
+    @Suppress("DEPRECATION")
+    defaultTracking: DefaultTrackingOptions = DefaultTrackingOptions(),
```
Mercy811, line 44: why are we removing `@Deprecated` and adding `@Suppress` here? 
```suggestion
    @Deprecated("Please use 'autocapture.sessions' instead.", ReplaceWith("autocapture.sessions"))
    trackingSessionEvents: Boolean = true,
    @Deprecated("Please use 'autocapture' instead.", ReplaceWith("autocapture"))
    defaultTracking: DefaultTrackingOptions = DefaultTrackingOptions(),
```

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -39,9 +39,10 @@ open class Configuration @JvmOverloads constructor(
...
+    val autocapture: MutableSet<AutocaptureOption> = mutableSetOf(AutocaptureOption.SESSIONS),
```
Mercy811, line 45: As we discussed yesterday, shouldn't we use keep `autocapture` immutable as plugin installation is only done once?

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -76,4 +77,39 @@ open class Configuration @JvmOverloads constructor(
     companion object {
         const val MIN_TIME_BETWEEN_SESSIONS_MILLIS: Long = 300000
     }
+
+    @Deprecated("Please use 'autocapture.sessions' instead.", ReplaceWith("autocapture.sessions"))
+    var trackingSessionEvents: Boolean
+        get() = AutocaptureOption.SESSIONS in autocapture
+        set(value) {
+            if (value) autocapture += AutocaptureOption.SESSIONS else autocapture -= AutocaptureOption.SESSIONS
+        }
```
Mercy811, line 86: Does this snippet of code disable session tracking with the following configuration?
```
val amplitude = Amplitude(
  Configuration(
    apiKey = AMPLITUDE_API_KEY,
    context = applicationContext,
    trackingSessionEvents = false,
    autocapture = true,
  )
)
```

IMO, the new config `autocapture` should take precedence over old configs `trackingSessionEvents` and `defaultTracking`.

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -76,4 +77,39 @@ open class Configuration @JvmOverloads constructor(
...
+
+    @Suppress("DEPRECATION")
```
Mercy811, line 88: Why do we suppress deprecation here?

android/src/main/java/com/amplitude/android/DefaultTrackingOptions.kt
```diff
@@ -1,20 +1,19 @@
...
+    screenViews: Boolean = false,
 ) {
-    var userInteractions = false
-        @ExperimentalAmplitudeFeature
-        set
-
     // Prebuilt options for easier usage
     companion object {
         @JvmField
+        @Deprecated("Use AutocaptureOptions instead", ReplaceWith("AutocaptureOptions(appLifecycles = true, deepLinks = true, screenViews = true, elementInteractions = true)"))
```
Mercy811, line 16: `DefaultTrackingOptions.ALL` should not set `elementInteractions = true`

android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
Mercy811: > Use one canonical representation of the options internally, probably the auto capture variant. Everything else should bridge to that.

Hi @crleona, I'm leaning more towards a different way of deprecation 
- Keep all old interface (`defaultTracking` and `trackingSessionEvents`) reference unchanged
- Pass `autocapture` value to old interfaces
   - if `autocapture.sessions` is set, set `defaultTracking` and `trackingSessionEvents` accordingly 
   - if `autocapture` is not set, check `defaultTracking` and set `trackingSessionEvents`
It also keeps backward compatible.

android/src/main/java/com/amplitude/android/Configuration.kt
PouriaAmini: b4c09984d035a3b408719a2e06c3d42eb4c2c01a should make this mutable.

android/src/main/java/com/amplitude/android/Configuration.kt
PouriaAmini: If any of the `trackingSessionEvents`, `defaultTracking.session`, or `AutocaptureOptions.SESSIONS in autocapture` is false, this means that one of them is set to false manually. So I think we should disable sessions in that case.

android/src/main/java/com/amplitude/android/Configuration.kt
PouriaAmini: This was to suppress the warning that defaultTracking is deprecated. This is removed on the new logic which is much easier to maintain. Please see b4c09984d035a3b408719a2e06c3d42eb4c2c01a.

android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
PouriaAmini: I implemented a new logic to discard any changes to the `defaultTracking` options and `trackingSessionsEvents` after the configuration is initialized. This is similar to how we're approaching default tracking deprecation in the Swift SDK. Please take a look at b4c09984d035a3b408719a2e06c3d42eb4c2c01a.

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -155,5 +155,43 @@ open class Amplitude(
 fun Amplitude(apiKey: String, context: Context, configs: Configuration.() -> Unit): com.amplitude.android.Amplitude {
     val config = Configuration(apiKey, context)
     configs.invoke(config)
-    return com.amplitude.android.Amplitude(config)
+    @Suppress("DEPRECATION")
+    val modifiedConfig = Configuration(
```
PouriaAmini, line 159: Any changes made by the function passed to this function to add defaultTracking options will be ignored unless a new `Configuration` is created from the modified configuration.

android/src/main/java/com/amplitude/android/DefaultTrackingOptions.kt
```diff
@@ -32,15 +32,4 @@ constructor(
                 screenViews = false,
             )
     }
-
-    @ExperimentalAmplitudeFeature
-    constructor(
```
yuhao900914, line 37: Is this backward compatible? 
The original way is not allowed anymore.
```DefaultTrackingOptions(
        sessions = true,
        appLifecycles = true,
        deepLinks = false,
        screenViews = true,
        userInteractions = true
    )
```

android/src/main/java/com/amplitude/android/Timeline.kt
```diff
@@ -138,11 +138,9 @@ class Timeline(
         val sessionEvents = mutableListOf<BaseEvent>()
         val configuration = amplitude.configuration as Configuration
         // If any trackingSessionEvents is false (default value is true), means it is manually set
-        @Suppress("DEPRECATION")
```
yuhao900914, line 141: val trackingSessionEvents = AutocaptureOption.SESSIONS in configuration.autocapture

android/src/test/java/com/amplitude/android/AmplitudeSessionTest.kt
```diff
@@ -64,7 +64,7 @@ class AmplitudeSessionTest {
         amplitudeDispatcherField.set(amplitude, dispatcher)
     }
 
-    private fun createConfiguration(storageProvider: StorageProvider? = null): Configuration {
```
yuhao900914, line 67: Should we keep the old test for backward compatibility and create new tests around auto capture for the new features?

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -39,9 +40,14 @@ open class Configuration @JvmOverloads constructor(
...
     var minTimeBetweenSessionsMillis: Long = MIN_TIME_BETWEEN_SESSIONS_MILLIS,
-    @Deprecated("Please use 'defaultTracking.sessions' instead.", ReplaceWith("defaultTracking.sessions"))
+    @Deprecated("Please use 'autocapture' instead and set 'AutocaptureOptions.SESSIONS' to enable the option.")
+    /** The SDK will no longer track changes to `trackingSessionEvents`'s value after initialization. */
     var trackingSessionEvents: Boolean = true,
+    @Suppress("DEPRECATION")
+    @Deprecated("Please use 'autocapture' instead", ReplaceWith("autocapture"))
+    /** The SDK will no longer track changes to the `defaultTracking` options after initialization. */
     var defaultTracking: DefaultTrackingOptions = DefaultTrackingOptions(),
+    autocapture: Set<AutocaptureOption> = setOf(AutocaptureOption.SESSIONS),
```
yuhao900914, line 50: should we also support boolean, making autocapture = true enable all autocapture options?

android/src/main/java/com/amplitude/android/DefaultTrackingOptions.kt
PouriaAmini: `userInteractions` has never been released. So I think it's safe to remove it entirely from `DefaultTracking` since no one's using it.

android/src/main/java/com/amplitude/android/Timeline.kt
PouriaAmini: I think I just have to remove the comment on line 140. we no longer need `trackingSessionEvents` since I'm using `AutocaptureOption.SESSIONS in configuration.autocapture` directly.

android/src/test/java/com/amplitude/android/AmplitudeSessionTest.kt
PouriaAmini: The old tests don't work since we're discarding `defaultTracking` changes after initialization. I can try to duplicate the configuration setup to test for `defaultTracking` and `trackingSessionEvents`.

android/src/main/java/com/amplitude/android/Configuration.kt
PouriaAmini: Based on the discussion we had on autocapture sync, having an ALL configuration for autocapture would make it hard for us to add new options later without surprising the customers.

android/src/main/java/com/amplitude/android/Timeline.kt
yuhao900914: I mean give assign AutocaptureOption.SESSIONS in configuration.autocapture to a variable.

android/src/test/java/com/amplitude/android/AmplitudeSessionTest.kt
yuhao900914: The old test shouldn't work. But it's still testing the default event tracking, so I think we shouldn't remove those test. Instead, we need to re-initialize Amplitude in every single test.

android/src/main/java/com/amplitude/android/Configuration.kt
yuhao900914: Good to know thanks.

android/src/main/java/com/amplitude/android/Timeline.kt
PouriaAmini: May I ask why we would want to assign it to a variable? Can't we use it directly?

yuhao900914: Just for simplifying the code. it makes the if statement too long.

android/src/test/java/com/amplitude/android/AmplitudeSessionTest.kt
yuhao900914: Are those tests for testing the default event tracking behavior or test the set config after initialization? If it's for the former reason, then we shouldn't remove those.

PouriaAmini: That test was for testing the default event tracking. I added it back in fc3ac6f1384c2ab0402549d419c73173ea96097c.

android/src/main/java/com/amplitude/android/DefaultTrackingOptions.kt
```diff
@@ -8,13 +10,10 @@ constructor(
...
     var screenViews: Boolean = false,
 ) {
-    var userInteractions = false
-        @ExperimentalAmplitudeFeature
-        set
-
     // Prebuilt options for easier usage
     companion object {
         @JvmField
+        @Deprecated("Use AutocaptureOptions instead", ReplaceWith("AutocaptureOptions(appLifecycles = true, deepLinks = true, screenViews = true)"))
```
Mercy811, line 16: It's better to explicitly list all DET configs here even though it's true by default
```suggestion
        @Deprecated("Use AutocaptureOptions instead", ReplaceWith("AutocaptureOptions(appLifecycles = true, deepLinks = true, screenViews = true, sessions = true)"))
```

android/src/main/java/com/amplitude/android/DefaultTrackingOptions.kt
```diff
@@ -24,6 +23,7 @@ constructor(
             )
 
         @JvmField
+        @Deprecated("Use AutocaptureOptions instead", ReplaceWith("AutocaptureOptions(sessions = false)"))
```
Mercy811, line 26: The same as above. Better to have all 
```suggestion
        @Deprecated("Use AutocaptureOptions instead", ReplaceWith("AutocaptureOptions(appLifecycles = false, deepLinks = false, screenViews = false, sessions = false)"))
```

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -76,4 +80,24 @@ open class Configuration @JvmOverloads constructor(
     companion object {
         const val MIN_TIME_BETWEEN_SESSIONS_MILLIS: Long = 300000
     }
+
+    val autocapture: Set<AutocaptureOption> = autocapture
+        @Suppress("DEPRECATION")
+        get() = autocaptureOptions {
+            if (trackingSessionEvents && defaultTracking.sessions && AutocaptureOption.SESSIONS in field) {
+                +sessions
+            }
```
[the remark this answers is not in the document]
PouriaAmini: `trackingSessionEvents` is default to true. If either of them is false, it means that it was changes manually and we shouldn't track sessions.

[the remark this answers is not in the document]
Mercy811: I confirm `trackingSessionEvents` is true by default at https://github.com/amplitude/Amplitude-Kotlin/blob/4b834bd366b6cef05d362973cc03075a48e7b1aa/android/src/main/java/com/amplitude/android/Configuration.kt#L43. Could you correct the docs later? 
image

android/src/main/java/com/amplitude/android/Configuration.kt
```diff
@@ -76,4 +80,24 @@ open class Configuration @JvmOverloads constructor(
...
+            if (defaultTracking.deepLinks || AutocaptureOption.DEEP_LINKS in field) {
+                +deepLinks
+            }
+            if (defaultTracking.screenViews || AutocaptureOption.SCREEN_VIEWS in field) {
+                +screenViews
+            }
+            if (AutocaptureOption.ELEMENT_INTERACTIONS in field) {
+                +elementInteractions
+            }
+        }
```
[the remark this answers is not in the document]
PouriaAmini: I think during migration, both autocapture and defaultTracking might be used. Also, discarding anything from `defaultTracking` if `autocapture` is set seems to be a coupling between the two configs, which can cause maintainability issues later.

android/src/main/java/com/amplitude/android/DefaultTrackingOptions.kt
PouriaAmini: Since `AutocaptureOption` is not an exact replacement for `DefaultTrackingOptions`, I removed `ReplaceWith`  entirely.

android/src/main/java/com/amplitude/android/Configuration.kt
Mercy811: Close this and will go as we discussed here https://amplitude.slack.com/archives/C04HSDQA2VB/p1722893976133219?thread_ts=1722556642.263359&cid=C04HSDQA2VB

android/src/main/java/com/amplitude/android/Configuration.kt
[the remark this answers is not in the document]
Mercy811: Close this and will go as we discussed here https://amplitude.slack.com/archives/C04HSDQA2VB/p1722893976133219?thread_ts=1722556642.263359&cid=C04HSDQA2VB

Mercy811 (commented):
Overall looks good to me. Please add a test for the constructor with `defaultTracking`.

Mercy811 (approved):
LGTM. Note that the constructors of configuration in both Swift and Kotlin SDK allow either defaultTracking or autocapture. The only difference is that
we can turn session & app lifecycle on/off after initialization in Kotlin Swift but can NOT do that in Swift. @PouriaAmini please make sure it's documented.

crleona (approved):

Issue #208, opened 2024-08-05
App Crashes with  java.lang.OutOfMemoryError: Failed to allocate a 32 byte allocation with 672880 free bytes and 657KB until OOM


BalaDakshina:
## Expected Behavior
App Should not crash when using Kotlin version of Amplitude SDK.

## Current Behavior
App Crashes due to amplitude SDK running out of memory

## Steps to Reproduce
1. integrate the kotlin version of the SDK
2. Run the app with invalid/fake apiKey.
3. Error Log before crash: 

```Event storage file not found: /data/user/0/com.ford.fcsdriverinterface.mock/app_amplitude-disk-queue/$default_instance-15-1722867019549-447: open failed: ENOENT (No such file or directory)
Event storage file not found: /data/user/0/com.ford.fcsdriverinterface.mock/app_amplitude-disk-queue/$default_instance-15-1722867019549-447: open failed: ENOENT (No such file or directory)
Event storage file not found: /data/user/0/com.ford.fcsdriverinterface.mock/app_amplitude-disk-queue/$default_instance-15-1722867019549-447: open failed: ENOENT (No such file or directory)
Google Play Services SDK not found for app set id!
user id and device id are null; amplitude may not resolve identity
Event storage file not found: /data/user/0/com.ford.fcsdriverinterface.mock/app_amplitude-disk-queue/$default_instance-15-1722867020387-898: open failed: ENOENT (No such file or directory)
Event storage file not found: /data/user/0/com.ford.fcsdriverinterface.mock/app_amplitude-disk-queue/$default_instance-15-1722867020727-257: open failed: ENOENT (No such file or directory)
Event storage file not found: /data/user/0/com.ford.fcsdriverinterface.mock/app_amplitude-disk-queue/$default_instance-15-1722867021175-235: open failed: ENOENT (No such file or directory)
Event storage file not found: /data/user/0/com.ford.fcsdriverinterface.mock/app_amplitude-disk-queue/$default_instance-15-1722867020387-898: open failed: ENOENT (No such file or directory)
Event storage file not found: /data/user/0/com.ford.fcsdriverinterface.mock/app_amplitude-disk-queue/$default_instance-15-1722867020727-257: open failed: ENOENT (No such file or directory)
Event storage file not found: /data/user/0/com.ford.fcsdriverinterface.mock/app_amplitude-disk-queue/$default_instance-15-1722867021175-235: open failed: ENOENT (No such file or directory)
[...]

PR #210, opened 2024-08-08
fix: Fix OOM error when a bad api key is passed

izaaz:
### Summary

When the server returns a 400 because of a bad api key, we continuously retry without dropping the data. This fixes that issue

Fixes issue #208

### Checklist


Mercy811 (approved):
Thank you Izaaz for fixing it. LGTM. Nits: add a test for in memory response handler

core/src/main/java/com/amplitude/core/utilities/FileResponseHandler.kt
```diff
@@ -65,7 +65,7 @@ class FileResponseHandler(
             removeCallbackByInsertId(eventsString)
             throw e
         }
-        if (eventsList.size == 1) {
+        if (eventsList.size == 1 || badRequestResponse.isInvalidApiKeyResponse()) {
```
yuhao900914, line 68: Here we remove the eventFile from the storage. 
Should we keep the event for an invalid API key instead?

izaaz: You mean keep the event for a valid future API key?

yuhao900914: Yes, keep it until the next try.  I saw [previous android SDK ](https://github.com/amplitude/Amplitude-Android/blob/main/src/main/java/com/amplitude/api/AmplitudeClient.java#L2337)just logged an error with invalid API.

crleona: The issue is that we will continuously retry requests that are known to be bad, there's no case in which the instance can recover until restarted with a new API key, which will also effectively orphan these events as storage is also segmented by api key.

yuhao900914: That makes sense. Thanks.

yuhao900914 (approved):

PR #216, opened 2024-08-12
feat: Capture top-level Fragments with Autocapture

PouriaAmini:
### Summary

This PR adds support for capturing top-level Fragments with the `screenView` configuration of Autocapture.

#### New Event
- `[Amplitude] Fragment Viewed`, this event is fired when the Fragment becomes visible on the screen.

#### New properties
- `[Amplitude] Fragment Class`, the canonical class name of the Fragment.
- `[Amplitude] Fragment Identifier`, the entry name of the fragment based on its internal id. The entry name is the id specified in the layout.
-  `[Amplitude] Screen Name`
- `[Amplitude] Fragment Tag`, the tag attached to the fragment.

### Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -63,7 +63,9 @@ dependencies {
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.2'
     implementation 'com.amplitude:analytics-connector:1.0.0'
     implementation 'androidx.core:core-ktx:1.7.0'
-    compileOnly 'androidx.compose.ui:ui:1.6.7'
+    implementation 'androidx.activity:activity-ktx:1.9.1'
+    compileOnly 'androidx.fragment:fragment-ktx:1.8.2'
+    compileOnly 'androidx.compose.ui:ui:1.6.8'
 
     coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'
 
--- a/android/src/main/java/com/amplitude/android/internal/fragments/AutocaptureFragmentLifecycleCallbacks.kt
+++ b/android/src/main/java/com/amplitude/android/internal/fragments/AutocaptureFragmentLifecycleCallbacks.kt
@@ -7,8 +7,8 @@ import com.amplitude.android.utilities.DefaultEventUtils.EventProperties
 import com.amplitude.android.utilities.DefaultEventUtils.EventTypes
 import com.amplitude.common.Logger
 
-class AutocaptureFragmentLifecycleCallbacks(
-    private val track: (String, Map<String, Any?>) -> Unit,
+internal class AutocaptureFragmentLifecycleCallbacks(
+    private val track: TrackEventCallback,
     private val logger: Logger
 ) : FragmentManager.FragmentLifecycleCallbacks() {
     override fun onFragmentResumed(fm: FragmentManager, f: Fragment) {
--- a/android/src/main/java/com/amplitude/android/internal/fragments/AutocaptureFragmentLifecycleCallbacks.kt
+++ b/android/src/main/java/com/amplitude/android/internal/fragments/AutocaptureFragmentLifecycleCallbacks.kt
@@ -1,5 +1,7 @@
 package com.amplitude.android.internal.fragments
 
+import android.os.Build
+import androidx.annotation.RequiresApi
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentManager
 import com.amplitude.android.utilities.DefaultEventUtils.Companion.screenName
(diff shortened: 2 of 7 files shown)


android/build.gradle
```diff
@@ -63,7 +63,9 @@ dependencies {
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.2'
     implementation 'com.amplitude:analytics-connector:1.0.0'
     implementation 'androidx.core:core-ktx:1.7.0'
-    compileOnly 'androidx.compose.ui:ui:1.6.7'
+    implementation 'androidx.activity:activity-ktx:1.9.1'
+    compileOnly 'androidx.fragment:fragment-ktx:1.8.2'
```
izaaz, line 67: For compose we check if compose is available or not. For this one, how do we make sure it is available before calling any of the classes inside this class?

PouriaAmini: We're checking if the `FragmentActivity` class is available [here](https://github.com/amplitude/Amplitude-Kotlin/pull/216/files#diff-115c82866176416f03f37c21dea752145c553bba0d00bbc905db9e630af6225bR59) before using it.

android/src/main/java/com/amplitude/android/internal/fragments/FragmentActivityHandler.kt
```diff
@@ -0,0 +1,32 @@
...
+import android.app.Activity
+import androidx.fragment.app.FragmentActivity
+import com.amplitude.common.Logger
+import java.util.WeakHashMap
+
+internal typealias TrackEventCallback = (String, Map<String, Any?>) -> Unit
+
+internal object FragmentActivityHandler {
+    private val callbacksMap =
+        WeakHashMap<FragmentActivity, AutocaptureFragmentLifecycleCallbacks>()
```
PouriaAmini, line 12: This approach will not work if there are multiple instances of Amplitude (the last callback will overwrite the previous one).

android/src/main/java/com/amplitude/android/internal/fragments/AutocaptureFragmentLifecycleCallbacks.kt
```diff
@@ -0,0 +1,45 @@
...
+
+        track(
+            EventTypes.FRAGMENT_VIEWED,
+            mapOf(
+                EventProperties.FRAGMENT_CLASS to className,
+                EventProperties.FRAGMENT_IDENTIFIER to fragmentIdentifier,
+                EventProperties.SCREEN_NAME to screenName,
+                EventProperties.FRAGMENT_TAG to fragmentTag,
+            )
+        )
```
Mercy811, line 43: Could you update the PR description to add
- which event properties are tracked with in the new event `[Amplitude] Fragment Viewed`
- if event properties can be null and when

android/src/main/java/com/amplitude/android/internal/fragments/FragmentActivityHandler.kt
```diff
@@ -0,0 +1,34 @@
...
+
+internal typealias TrackEventCallback = (String, Map<String, Any?>) -> Unit
+
+@RequiresApi(Build.VERSION_CODES.LOLLIPOP)
+internal object FragmentActivityHandler {
+    private val callbacksMap =
+        WeakHashMap<FragmentActivity, MutableList<AutocaptureFragmentLifecycleCallbacks>>()
+
+    fun Activity.registerFragmentLifecycleCallbacks(track: TrackEventCallback, logger: Logger) {
+        (this as? FragmentActivity)?.apply {
```
Mercy811, line 18: Could you handle the case where type cast from Activity to FragmentActivity fails? 

If type cast fails, what will happen?

android/src/main/java/com/amplitude/android/internal/fragments/FragmentActivityHandler.kt
```diff
@@ -0,0 +1,34 @@
...
+            val callback = AutocaptureFragmentLifecycleCallbacks(track, logger)
+            supportFragmentManager.registerFragmentLifecycleCallbacks(callback, false)
+            callbacksMap.getOrPut(this) { mutableListOf() }.add(callback)
+        }
+    }
+
+    fun Activity.unregisterFragmentLifecycleCallbacks() {
+        (this as? FragmentActivity)?.apply {
```
Mercy811, line 26: Type cast failure handle +1

Mercy811 (commented):
Thanks Pouria for adding the support for fragment tracking. See comments

android/src/main/java/com/amplitude/android/internal/fragments/FragmentActivityHandler.kt
PouriaAmini: If the cast fails, the activity is not a `FragmentActivity` and does not contain any fragments.

PouriaAmini: Please let me know if there is any specific behavior that we want for the failure case.

Mercy811: I was thinking if we should track when type cast fails or at least log out for debugging, for example, fragment tracking is not working because the type cast failure.

Mercy811 (approved):
Thank you Pouria, LGTM. Could you add logs to when type cast fails? Thank you

PR #227, opened 2024-10-04
fix: don't delete old profile file for backwards compatibility

izaaz:
### Summary

Don't delete the old profile file since we are not doing a major version upgrade. Deleting this file prevents will cause issues when a customer rolls back the Amplitude SDK to an older version or rollsback their app.

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/migration/IdentityStorageMigration.kt
+++ b/android/src/main/java/com/amplitude/android/migration/IdentityStorageMigration.kt
@@ -18,7 +18,9 @@ class IdentityStorageMigration(
             if (identity.deviceId != null) {
                 destination.saveDeviceId(identity.deviceId)
             }
-            source.delete()
+            // Since we are not doing a major version upgrade, keep the old profile file around
+            // just in case someone decides to rollback back to the older version of the SDK.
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/amplitude/android/migration/IdentityStorageMigration.kt
```diff
@@ -18,7 +18,9 @@ class IdentityStorageMigration(
             if (identity.deviceId != null) {
                 destination.saveDeviceId(identity.deviceId)
             }
-            source.delete()
+            // Since we are not doing a major version upgrade, keep the old profile file around
+            // just in case someone decides to rollback back to the older version of the SDK.
+//            source.delete()
```
crleona, line 23: not a priority, but is there a cleanup function that would delete this?

crleona (approved):

android/src/main/java/com/amplitude/android/migration/IdentityStorageMigration.kt
izaaz: Synced up offline. We currently don't have a cleanup function and we'll prolly create one in the future

PR #232, opened 2025-01-07
fix: autocapture missing events fix

igor-amp:
### Summary

Autocapture fix for missing and broken event sequence

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
+++ b/android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
@@ -8,12 +8,18 @@ import android.os.Bundle
 import com.amplitude.android.AutocaptureOption
 import com.amplitude.android.Configuration
 import com.amplitude.android.ExperimentalAmplitudeFeature
+import com.amplitude.android.utilities.ActivityCallbackType
+import com.amplitude.android.utilities.ActivityLifecycleObserver
 import com.amplitude.android.utilities.DefaultEventUtils
 import com.amplitude.core.Amplitude
 import com.amplitude.core.platform.Plugin
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
 import com.amplitude.android.Amplitude as AndroidAmplitude
 
-class AndroidLifecyclePlugin : Application.ActivityLifecycleCallbacks, Plugin {
+class AndroidLifecyclePlugin(
+    private val activityLifecycleObserver: ActivityLifecycleObserver
+) : Application.ActivityLifecycleCallbacks, Plugin {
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -57,26 +63,35 @@ class AndroidLifecyclePlugin : Application.ActivityLifecycleCallbacks, Plugin {
     }
 
     override fun onActivityStarted(activity: Activity) {
+        started.add(activity.hashCode())
+
+        if (AutocaptureOption.APP_LIFECYCLES in androidConfiguration.autocapture && started.size == 1) {
+            DefaultEventUtils(androidAmplitude).trackAppOpenedEvent(
```
izaaz, line 69: Why don't we track app opened with app updated/installed events?

igor-amp: Application installation does not always follow with application open/close

igor-amp: and context can be initialized with foreground/background service

izaaz (approved):

izaaz (approved):
Interesting!

crleona (approved):

android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -5,58 +5,101 @@ import android.app.Application
...
-import java.util.concurrent.atomic.AtomicBoolean
-import java.util.concurrent.atomic.AtomicInteger
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import com.amplitude.android.Amplitude as AndroidAmplitude
 
-class AndroidLifecyclePlugin : Application.ActivityLifecycleCallbacks, Plugin {
+class AndroidLifecyclePlugin(
+    private val activityLifecycleObserver: ActivityLifecycleObserver
+) : Application.ActivityLifecycleCallbacks, Plugin {
```
izaaz, line 23: Does this need to extend ActivityLifecycleCallbacks anymore?

izaaz (approved):

android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
igor-amp: technically no, but this solution is temporary and removing interface will also force removing @Override annotations and nothing more. Right now everything is easy to reverse in this way.

PR #237, opened 2025-02-05
feat: cleaning up http client interface and adding the ability to use custom http clients

izaaz:
### Summary
1. Moving http related classes into a package
2. Refactoring the http client class to expose a single method called upload
3. Creating an interface
4. Testing it out with an OkHttp custom client

### Checklist


core/src/main/java/com/amplitude/core/Configuration.kt
```diff
@@ -54,6 +56,16 @@ open class Configuration @JvmOverloads constructor(
             true
         }
     }
+
+    internal fun getApiHost(): String {
```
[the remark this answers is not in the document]
igor-amp: copy paste and format it please

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
```diff
@@ -134,15 +135,14 @@ class EventPipeline(
...
-                        val connection = httpClient.upload()
-                        connection.outputStream?.let {
-                            connection.setEvents(eventsString)
-                            connection.setDiagnostics(amplitude.diagnostics)
-                            // Upload the payloads.
-                            connection.close()
-                        }
 
-                        responseHandler.handle(connection.response, events, eventsString)
+                        val diagnostics = if (amplitude.diagnostics.hasDiagnostics()) {
```
igor-amp, line 139: in source code, `hasDiagnostics()` is imbedded into `extractDiagnostics()`. No need to check here. But maybe worth returning null from `extractDiagnostics()`.

igor-amp (approved):
lgtm

samples/kotlin-android-app/src/main/java/com/amplitude/android/sample/CustomOkHttpClient.kt
```diff
@@ -0,0 +1,31 @@
...
+
+class CustomOkHttpClient(val apiKey: String) : HttpClientInterface {
+    private val okHttpClient = OkHttpClient()
+
+    override fun upload(events: String, diagnostics: String?): AnalyticsResponse {
+        val mediaType = "application/json; charset=utf-8".toMediaTypeOrNull()
+        val ampRequest = AnalyticsRequest(apiKey, events, diagnostics = diagnostics)
+        val formBody: RequestBody = RequestBody.create(mediaType, ampRequest.getBodyStr())
+        val request: Request =
+            Request.Builder().url("https://api.amplitude.com/2/httpapi").post(formBody).build()
```
cbeyls, line 20: How are customers supposed to implement a custom API client properly without having access to `getServerUrl()` or `minIdLength` from the internal configuration? Hardcoding these values is definitely not a good option.

izaaz: The initial plan was to make our http client an open class which can be extended. But we were running into a circular dependency.

The http client needs a configuration to be initialized and the custom http client needed to be passed into the configuration so that the Amplitude client can be configured to use a custom http client. 

Since the HttpClient is an interface, you can define your class that takes in the Amplitude configuration to fetch all the properties from the configuration directly.

For example

```
class CustomClient() : HttpClientInterface {
    fun initialize(configuration: Configuration) {
        ....
    }

    // override upload here 
}
```

And in init

```
val customClient = CustomClient()
val config = Configuration(...)
customClient.initialize(config)
val amplitude = Amplitude(configuration)
```

I can update the example to show this and also open to suggestions.

izaaz: Something like this: https://github.com/amplitude/Amplitude-Kotlin/pull/240

Issue #114, opened 2023-03-30
KMM support for iOS


TSkovsgaard:
## Summary

Hi I'm working on a Kotlin Multiplatform project, is this SDK intended to be extended to support iOS/Apple?

posplaw:
Second this

salexwm:
4th this

nathanfallet:
I was thinking about making it myself because it seems not to be planned yet.

nathanfallet:
Made the first version of the `amplitude-kmp` sdk here: https://github.com/guimauvedigital/amplitude-kmp.

I works fine on Android, but I still have an issue on iOS and I haven't been able to figure it out yet: https://github.com/guimauvedigital/amplitude-kmp/issues/1

DroidPulkit:
We still don't have official KMP support by Amplitude team?
It's almost 3 years, since the start of this issue. So, far I see 26 people thumbs up this feature request.

nathanfallet:
@DroidPulkit That's why I made the amplitude-kmp lib on my own (see link above)

PR #241, opened 2025-02-22
refactor: preparing gradle configuration to support kmm

nathanfallet:
### Summary

As we can see in #114, a lot of people (including me) would love support for KMM support in Amplitude-Kotlin.
That's why I started to prepare the repository to add this support. Here's what changed and why:
- Rewriting `build.gradle` configs for code that can be shared, using multiplatform plugin instead of jvm one
- Renaming `src/main/java` to `src/commonMain/kotlin` for those same modules, so that we all continue with this same path allowing the library to continue to evolve while we start updating the code (avoid merge conflicts later)

If this gets approved and merged, I'll continue updating the common/core modules to make them multiplatform compatible, and work on an ios module later on.

### Checklist


nathanfallet:
I'm creating a new KMP app today, and I can see the team has not changed anything about this (no review, plan or whatever). Will have to stay with the old manual way...

nathanfallet:
I'm closing this as it is outdated. As mentioned in #114, I'll work on a wrapper library to make this work.

PR #252, opened 2025-03-06
fix: respect play services flags inside context provider

polbins:
- this should now respect the value set from tracking options

### Summary

Adds `shouldTrackAppSetId` to context provider, this respects the value set from tracking options.

### Checklist

### DIFF
--- a/common-android/src/main/java/com/amplitude/common/android/AndroidContextProvider.kt
+++ b/common-android/src/main/java/com/amplitude/common/android/AndroidContextProvider.kt
@@ -20,7 +20,8 @@ import java.util.Locale
 class AndroidContextProvider(
     private val context: Context,
     private val locationListening: Boolean,
-    private val shouldTrackAdid: Boolean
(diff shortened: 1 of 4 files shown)


common-android/src/main/java/com/amplitude/common/android/AndroidContextProvider.kt
```diff
@@ -194,6 +195,10 @@ class AndroidContextProvider(
         }
 
         private fun fetchAppSetId(): String? {
+            if (!shouldTrackAppSetId) {
```
igor-amp, line 198: This function both set the `appSetId` internally and return it to set it again inside of init block. Let's set it only once.
This will allow to remove if(not true) check.

polbins: Good catch! That's a weird pattern, I'll update it for sure 👍🏽

igor-amp (approved):

PR #253, opened 2025-03-12
fix: ANR fix attempt on network change

polbins:
### Summary

This is an attempt to fix an ANR: it seems that the stack trace point to the call on `amplitude.flush()` from `AndroidNetworkConnectivityCheckerPlugin` which happens Activity onCreate. 

image

In this PR, we ensure that flush is uniformly launched on the Amplitude dispatcher.

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -119,11 +119,8 @@ open class Amplitude(
         dummyExitForegroundEvent.timestamp = timestamp
         timeline.process(dummyExitForegroundEvent)
 
-        amplitudeScope.launch(amplitudeDispatcher) {
-            isBuilt.await()
-            if ((configuration as Configuration).flushEventsOnClose) {
-                flush()
-            }
+        if ((configuration as Configuration).flushEventsOnClose) {
+            flush()
         }
     }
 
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/amplitude/android/plugins/AndroidNetworkConnectivityCheckerPlugin.kt
```diff
@@ -29,7 +29,10 @@ class AndroidNetworkConnectivityCheckerPlugin : Plugin {
                 override fun onNetworkAvailable() {
                     amplitude.logger.debug("AndroidNetworkListener, onNetworkAvailable.")
                     amplitude.configuration.offline = false
-                    amplitude.flush()
+
+                    amplitude.amplitudeScope.launch(amplitude.amplitudeDispatcher) {
```
izaaz, line 33: Is this the rootcause or one of the problems?

polbins: For one of them, yes. From SDK Console stack traces:

image

polbins: This on network change call back is called upon onCreate when I tested locally

igor-amp:
To verify that this change makes difference let's block ConnectivityThread and get ANR log.

igor-amp (approved):
lgtm

PR #254, opened 2025-03-13
fix: remove synchronized on mediator

polbins:
### Summary

Removes `synchronized` blocks on `Mediator` in favor of `CopyOnWriteArrayList` instead.

It would be good to review this PR on a per commit basis.

### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/platform/Mediator.kt
+++ b/core/src/main/java/com/amplitude/core/platform/Mediator.kt
@@ -5,7 +5,7 @@ import com.amplitude.core.events.GroupIdentifyEvent
 import com.amplitude.core.events.IdentifyEvent
 import com.amplitude.core.events.RevenueEvent
 
-internal class Mediator(internal val plugins: MutableList<Plugin>) {
+internal class Mediator(private val plugins: MutableList<Plugin>) {
     fun add(plugin: Plugin) = synchronized(plugins) {
         plugins.add(plugin)
     }
@@ -58,4 +58,9 @@ internal class Mediator(internal val plugins: MutableList<Plugin>) {
             closure(it)
         }
     }
+
+    /**
+     * Only visible for testing
+     */
+    internal fun size() = plugins.size
 }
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/amplitude/core/platform/Mediator.kt
```diff
@@ -4,58 +4,55 @@ import com.amplitude.core.events.BaseEvent
...
         var result: BaseEvent? = event
 
         plugins.forEach { plugin ->
-            if (result != null) {
-                when (plugin) {
+            val eventToBeProcessed = result
+            if (eventToBeProcessed != null) {
+                result = when (plugin) {
                     is DestinationPlugin -> {
                         try {
```
igor-amp, line 24: It's not part of the change, but why we try catch only here? let's wrap the hole when with try catch to cover other executes

core/src/main/java/com/amplitude/core/platform/Timeline.kt
```diff
@@ -2,13 +2,14 @@ package com.amplitude.core.platform
...
 import com.amplitude.core.events.BaseEvent
+import java.util.concurrent.CopyOnWriteArrayList
 
 open class Timeline {
     internal val plugins: Map<Plugin.Type, Mediator> = mapOf(
-        Plugin.Type.Before to Mediator(mutableListOf()),
-        Plugin.Type.Enrichment to Mediator(mutableListOf()),
-        Plugin.Type.Destination to Mediator(mutableListOf()),
-        Plugin.Type.Utility to Mediator(mutableListOf())
+        Plugin.Type.Before to Mediator(CopyOnWriteArrayList()),
```
igor-amp, line 9: any reason we don't make it default in constructor?

igor-amp (changes requested):

core/src/main/java/com/amplitude/core/platform/Mediator.kt
```diff
@@ -4,58 +4,55 @@ import com.amplitude.core.events.BaseEvent
...
-    fun applyClosure(closure: (Plugin) -> Unit) = synchronized(plugins) {
+    fun applyClosure(closure: (Plugin) -> Unit) {
         plugins.forEach {
             closure(it)
         }
     }
+
+    /**
+     * Only visible for testing
+     */
```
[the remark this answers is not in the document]
polbins: I'd like that too, but that is an Android dependency :D

core/src/main/java/com/amplitude/core/platform/Timeline.kt
polbins: No strong reason, I just refactored this manually.
Let me make that the default. 👍🏽

igor-amp (approved):

PR #258, opened 2025-03-20
refactor: remove-previous-retry-logic

polbins:
### Summary

This is a pre-requisite change prior to implementing the updated retry logic.

The goal of this PR is to revert the previous logic that doesn't align with the updated one that we want. (e.g. we're not going to modify the flushInterval on EventPipeline, and put most of the retry logic inside the Pipeline itself)

### Checklist

### DIFF
--- a/core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
+++ b/core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
@@ -30,6 +30,15 @@ class EventPipeline(
     private var scheduled: Boolean
     var flushSizeDivider: AtomicInteger = AtomicInteger(1)
 
+    private val responseHandler by lazy {
+        storage.getResponseHandler(
+            this@EventPipeline,
+            amplitude.configuration,
+            scope,
+            amplitude.retryDispatcher,
+        )
+    }
+
     companion object {
         internal const val UPLOAD_SIG = "#!upload"
     }
@@ -116,7 +125,8 @@ class EventPipeline(
                         if (eventsString.isEmpty()) continue
 
                         val diagnostics = amplitude.diagnostics.extractDiagnostics()
-                        httpClient.upload(eventsString, diagnostics)
+                        val response = httpClient.upload(eventsString, diagnostics)
+                        responseHandler.handle(response, events, eventsString)
                     } catch (e: FileNotFoundException) {
                         e.message?.let {
                             amplitude.logger.warn("Event storage file not found: $it")
--- a/core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
+++ b/core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
@@ -4,7 +4,6 @@ import com.amplitude.core.Amplitude
 import com.amplitude.core.events.BaseEvent
 import com.amplitude.core.utilities.http.HttpClient
 import com.amplitude.core.utilities.http.HttpClientInterface
-import com.amplitude.core.utilities.http.ResponseHandler
 import com.amplitude.core.utilities.logWithStackTrace
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.Channel.Factory.UNLIMITED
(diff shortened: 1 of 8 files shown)


polbins:
Marking as draft, this should be merged together with the updated retry logic.
But feel free to add in your review if you'd like! It shouldn't change much.

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
```diff
@@ -20,35 +19,30 @@ class EventPipeline(
...
-    var flushInterval = amplitude.configuration.flushIntervalMillis.toLong()
-    var flushQueueSize = amplitude.configuration.flushQueueSize
-
     private var running: Boolean
-
     private var scheduled: Boolean
-
     var flushSizeDivider: AtomicInteger = AtomicInteger(1)
 
-    var exceededRetries = false
```
igor-amp, line 44: 👍

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
```diff
@@ -153,15 +139,15 @@ class EventPipeline(
...
 
     private fun getFlushCount(): Int {
-        val count = flushQueueSize / flushSizeDivider.get()
+        val count = amplitude.configuration.flushQueueSize / flushSizeDivider.get()
         return count.takeUnless { it == 0 } ?: 1
     }
 
     private fun schedule() =
         scope.launch(amplitude.storageIODispatcher) {
-            if (isActive && running && !scheduled && !exceededRetries) {
```
igor-amp, line 162: with this change is it still a refactor or a bug fix itself?

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -100,20 +99,27 @@ class EventsFileManager(
...
-        val fileList =
-            directory.listFiles { _, name ->
-                name.contains(storageKey) && !name.endsWith(".tmp") && !name.endsWith(".properties")
-            } ?: emptyArray()
-        return fileList.sortedBy { it ->
+        val fileList = directory.listFiles { _, name ->
+            name.contains(storageKey) && !name.endsWith(".tmp") && !name.endsWith(".properties")
+        } ?: emptyArray()
+
+        fun getSortKeyForFile(file: File): String {
```
igor-amp, line 106: why do we move this function here? it will be harder to test it now.

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
```diff
@@ -160,8 +166,8 @@ class EventsFileManager(
                 return@withLock ""
             }
             filePathSet.add(filePath)
-            File(filePath).bufferedReader().use<BufferedReader, String> {
-                val content = it.readText()
+            File(filePath).bufferedReader().use { reader ->
+                val content = reader.readText()
```
igor-amp, line 170: 👍

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
polbins: It's a clean-up refactor, bug fix will be in the next PR: https://github.com/amplitude/Amplitude-Kotlin/pull/259

core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
polbins: It's only used in this `read()` function, so I decided to move it here. It shouldn't affect testing as the entry point will still be `read()`. (you can also see `EventsFileManagerTest` has no changes here)

But I'm open to actually not moving this inside the `read()`, I think it's more of a preference thing. Do you want to leave it outside of `read()`?

igor-amp (approved):
Don't merge without: https://github.com/amplitude/Amplitude-Kotlin/pull/259

PR #259, opened 2025-03-21
fix: retry upload order

polbins:
### Summary

After removal of the old retry logic in https://github.com/amplitude/Amplitude-Kotlin/pull/258, we're now re-introducing an updated version of retrying in this PR.

We have an updated `AnalyticsResponse` with a flag to retry when upload failed. The premise is that we will retry when the event file is still present, and won't retry if it's not (e.g. bad event file).

After knowing the result of the response, we'll now **retry** within the `EventPipeline` itself through the use of `ExponentialBackoffRetryHandler`. This class contains the logic for exponential back-off delay that happens every retry attempt.

We've also updated how we're reading the event file from storage. We'll get the sorted event files again, whenever we're retrying.

### Checklist

### DIFF
--- a/android/src/test/java/com/amplitude/android/ResponseHandlerTest.kt
+++ b/android/src/test/java/com/amplitude/android/ResponseHandlerTest.kt
@@ -12,6 +12,10 @@ import io.mockk.every
 import io.mockk.mockkConstructor
 import io.mockk.mockkStatic
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.TestCoroutineScheduler
+import kotlinx.coroutines.test.advanceUntilIdle
+import kotlinx.coroutines.test.runTest
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
 import okhttp3.mockwebserver.RecordedRequest
@@ -26,12 +30,14 @@ import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 import java.util.concurrent.TimeUnit
 
+private const val FLUSH_INTERVAL_IN_MS = 150
+
+@ExperimentalCoroutinesApi
 @RunWith(RobolectricTestRunner::class)
 class ResponseHandlerTest {
     private lateinit var server: MockWebServer
     private lateinit var amplitude: Amplitude
 
-    @ExperimentalCoroutinesApi
     @Before
     fun setup() {
         server = MockWebServer()
@@ -46,7 +52,9 @@ class ResponseHandlerTest {
                 context = context,
                 serverUrl = server.url("/").toString(),
                 autocapture = setOf(),
-                flushIntervalMillis = 150,
+                // required for [PayloadTooLargeResponse] test
+                flushQueueSize = 2,
+                flushIntervalMillis = FLUSH_INTERVAL_IN_MS,
                 identifyBatchIntervalMillis = 1000,
                 flushMaxRetries = 3,
                 identityStorageProvider = IMIdentityStorageProvider(),
@@ -84,7 +92,8 @@ class ResponseHandlerTest {
     }
 
     @Test
-    fun `test handle on rate limit`() {
+    fun `test handle on rate limit`() = runTest {
+        setAmplitudeDispatchers(amplitude, testScheduler)
         val rateLimitBody = """
         {
           "code": 429,
@@ -95,18 +104,24 @@ class ResponseHandlerTest {
         for (i in 1..6) {
             server.enqueue(MockResponse().setBody(rateLimitBody).setResponseCode(429))
         }
+
+        amplitude.isBuilt.await()
         for (k in 1..4) {
             amplitude.track("test event $k")
-            runRequest()
+            advanceUntilIdle()
         }
         Thread.sleep(100)
+
         // verify the total request count when reaching max retries
         assertEquals(6, server.requestCount)
     }
 
     @Test
     fun `test handle payload too large with only one event`() {
-        server.enqueue(MockResponse().setBody("{\"code\": \"413\", \"error\": \"payload too large\"}").setResponseCode(413))
+        server.enqueue(
+            MockResponse().setBody("{\"code\": \"413\", \"error\": \"payload too large\"}")
+                .setResponseCode(413)
+        )
         val options = EventOptions()
         var statusCode = 0
         var callFinished = false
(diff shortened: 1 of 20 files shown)


core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
```diff
@@ -17,21 +21,24 @@ import java.util.concurrent.atomic.AtomicInteger
...
-    private val storage get() = amplitude.storage
-    private val scope get() = amplitude.amplitudeScope
+        ?: HttpClient(amplitude.configuration),
+    private val retryUploadHandler: ExponentialBackoffRetryHandler =
+        ExponentialBackoffRetryHandler(),
+    private val storage: Storage = amplitude.storage,
+    private val scope: CoroutineScope = amplitude.amplitudeScope,
+    private val writeChannel: Channel<WriteQueueMessage> = Channel(UNLIMITED),
+    private var uploadChannel: Channel<String> = Channel(UNLIMITED),
+    overrideResponseHandler: ResponseHandler? = null,
```
polbins, line 33: This is a hack that I needed to do so I can override the `ResponseHandler` for testing. I can't just initialize the `responseHandler` because it relies on the `EventPipeline` class itself (one of the cyclic dependency problems that we have in the SDK).

Let me know if there's a better alternative that we can use.

android/src/test/java/com/amplitude/android/ResponseHandlerTest.kt
```diff
@@ -287,4 +339,26 @@ class ResponseHandlerTest {
...
+
+    companion object {
+        private fun setAmplitudeDispatchers(
+            amplitude: com.amplitude.core.Amplitude,
+            testCoroutineScheduler: TestCoroutineScheduler,
+        ) {
+            // inject these dispatcher fields with reflection, as the field is val (read-only)
+            listOf(
+                "amplitudeDispatcher",
+                "networkIODispatcher",
```
igor-amp, line 351: networkIODispatcher

android/src/test/java/com/amplitude/android/ResponseHandlerTest.kt
```diff
@@ -287,4 +339,25 @@ class ResponseHandlerTest {
         every { anyConstructed<AndroidContextProvider>().mostRecentLocation } returns null
         every { anyConstructed<AndroidContextProvider>().appSetId } returns ""
     }
+
+    companion object {
+        private fun setAmplitudeDispatchers(
```
igor-amp, line 344: let's make one more Amplitude constructor that accepts dispatchers and context and set dispatchers directly in constructor. This will allow to take control over what is happening in init block while we have it and also removes this reflection from tests.

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
igor-amp: @polbins yea, please create a ticket for backlog to untangle this, it's ugly to have this spaghetti in our codebase.

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
```diff
@@ -118,25 +122,48 @@ class EventPipeline(
...
-                        e.logWithStackTrace(amplitude.logger, "Error when uploading event")
+    private suspend fun uploadNextEventFile() {
+        try {
+            // only get first event file, we want to upload them one by one, in order
+            val eventFile = storage.readEventsContent().firstOrNull() ?: return
+            val eventsString = storage.getEventsString(eventFile)
+            if (eventsString.isEmpty()) return
+
+            val diagnostics = amplitude.diagnostics.extractDiagnostics()
+            val response = httpClient.upload(eventsString, diagnostics)
```
igor-amp, line 137: I'm not sure I understand why we have so complex logic here:
1. looks like handled can be extracted from the `AnalyticsResponse`.
2. Do we expect that first file from the `storage.readEventsContent()` will change? Or do we expect that their order will change or the list size will change? why do we need to abandon the hole list of files?
3. Why do we send `UPLOAD_SIG` again if we can just call `httpClient.upload(eventsString, diagnostics)` again?
4. There is no need in recreation of a channel but even if we went this path it will not work.

igor-amp (changes requested):
That's a bit overcomplicated logic with a bug in reset.

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
polbins: As discussed in the meeting, I'll do a revamp of the changes here and let's do a re-review afterwards. Thanks for the review @igor-amp!

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
```diff
@@ -118,15 +125,36 @@ class EventPipeline(
...
+                        // if we encounter a retryable error, we retry with delay and
+                        // restart the loop to get the newest event files
+                        if (response.status.shouldRetryUploadOnFailure == true) {
+                            retryUploadHandler.attemptRetry { canRetry ->
+                                val retrySignal = if (canRetry) UPLOAD_SIG else MAX_RETRY_ATTEMPT_SIG
+                                uploadChannel.trySend(retrySignal)
+                            }
+                            break
+                        }
+                        retryUploadHandler.reset() // always reset when we've successfully uploaded
```
igor-amp, line 157: 👍

core/src/main/java/com/amplitude/core/utilities/FileResponseHandler.kt
```diff
@@ -38,12 +38,14 @@ class FileResponseHandler(
         try {
             eventsList = JSONArray(eventsString).toEvents()
         } catch (e: JSONException) {
-            storage.removeFile(eventFilePath)
+            scope.launch(storageDispatcher) {
```
igor-amp, line 41: Why do we make coroutines only for some operations and not the hole function?

core/src/main/java/com/amplitude/core/utilities/FileResponseHandler.kt
```diff
@@ -38,12 +38,14 @@ class FileResponseHandler(
...
+                storage.removeFile(eventFilePath)
+            }
             removeCallbackByInsertId(eventsString)
             throw e
         }
         triggerEventsCallback(eventsList, HttpStatus.SUCCESS.code, "Event sent success.")
-        scope.launch(dispatcher) {
+        scope.launch(storageDispatcher) {
             storage.removeFile(eventFilePath)
```
igor-amp, line 49: so, if you catch the error above you call `storage.removeFile(eventFilePath)` twice? is that a desired behavior?
Maybe use finally?

core/src/test/kotlin/com/amplitude/core/platform/EventPipelineTest.kt
```diff
@@ -5,82 +5,205 @@ import com.amplitude.core.Configuration
...
+        )
+
+        fakeResponse = PayloadTooLargeResponse(JSONObject())
+        val event = BaseEvent().apply { eventType = "test_event" }
+        eventPipeline.start()
+        eventPipeline.put(event)
+        advanceUntilIdle()
+
+        coVerify { retryUploadHandler.attemptRetry(any<(Boolean) -> Unit>()) }
+        // this will be called on the MAX_RETRY_ATTEMPT_SIG block on upload()
```
igor-amp, line 206: `retryUploadHandler.reset() // always reset when we've successfully uploaded` is called after every event is processed so can't be used to identify if `MAX_RETRY_ATTEMPT_SIG` was sent

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
```diff
@@ -118,15 +125,36 @@ class EventPipeline(
...
 
-                val eventsData = storage.readEventsContent()
-                for (events in eventsData) {
+                if (signal == MAX_RETRY_ATTEMPT_SIG) {
+                    amplitude.logger.debug(
+                        "Max retries ${retryUploadHandler.maxRetryAttempt} reached, temporarily stop consuming upload signals."
+                    )
+                    // approximately 32 seconds after max retry attempt is reached
+                    delay(retryUploadHandler.exponentialBackOffDelayInMs * 2)
+                    retryUploadHandler.reset()
```
igor-amp, line 134: we only reset here, do we wanna send upload signal again?

igor-amp (approved):
@polbins resolve comments and let's merge

core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
polbins: No, we don't need to as the processing will just continue from here after the delay.

core/src/main/java/com/amplitude/core/utilities/FileResponseHandler.kt
polbins: It will throw, so execution would stop. I think that behavior is fine right? So whichever case happens, we'll always call `removeFile`.

igor-amp: omg... why is it throwing? ok, let's clean it up later.

core/src/test/kotlin/com/amplitude/core/platform/EventPipelineTest.kt
polbins: This won't continue the processing because of how `InMemoryStorage` behaves during the test. 
- `eventPipeline.put(event)` will add 1 event to in memory storage's buffer
- `storage.readEventsContent()` will clear that event buffer

On the test, when retry signal is received and the second call to `readEventsContent()` is made, it won't continue the processing of the event. I added more comments to clarify this behavior in the test. It's a bit tricky, but it actually serves a good purpose for one-off kind of tests like this. (The retry behavior is tested via `ResponseHandlerTest` which is done through Robolectric)

PR #261, opened 2025-04-01
fix: compose autocapture properties update

igor-amp:
### Summary

Jira ticket: https://amplitude.atlassian.net/browse/AMP-126719
With Compose update to new version it's internal implementation of testTag changed from `SemanticsModifier` to `TestTagElement`
The only common part is Inspectable Value params that we can iterate.
Tested on versions:
from 1.5.+
to 1.7.+

Supported tags:
`Modifier.testTag("some value")`
`Modifier.semantics { testTag = "some value" }`
Clickable nodes added support for clickable InspectableValue.

### Checklist

* Does your PR have a breaking change?:   no

### DIFF
--- a/android/src/main/java/com/amplitude/android/internal/ViewHierarchyScanner.kt
+++ b/android/src/main/java/com/amplitude/android/internal/ViewHierarchyScanner.kt
@@ -72,21 +72,24 @@ internal object ViewHierarchyScanner {
             if (view is ViewGroup) {
                 queue.addAll(view.children)
             }
-
-            // Applies the locators until a target is found. If the target type is clickable, check
-            // the children in case the target is a child which is also clickable.
-            viewTargetLocators.any { locator ->
-                with(locator) {
-                    view.locate(targetPosition, targetType)?.let { newTarget ->
-                        if (targetType == ViewTarget.Type.Clickable) {
-                            target = newTarget
-                            return@any true
-                        } else {
-                            return newTarget
+            try {
+                // Applies the locators until a target is found. If the target type is clickable, check
+                // the children in case the target is a child which is also clickable.
+                viewTargetLocators.any { locator ->
+                    with(locator) {
+                        view.locate(targetPosition, targetType)?.let { newTarget ->
+                            if (targetType == ViewTarget.Type.Clickable) {
+                                target = newTarget
+                                return@any true
+                            } else {
+                                return newTarget
+                            }
                         }
+                        false
                     }
-                    false
                 }
+            } catch (t: Throwable) {
+                logger.error("Error while locating target")
             }
         }
 
--- a/android/src/main/java/com/amplitude/android/internal/ViewHierarchyScanner.kt
+++ b/android/src/main/java/com/amplitude/android/internal/ViewHierarchyScanner.kt
@@ -88,8 +88,8 @@ internal object ViewHierarchyScanner {
                         false
                     }
                 }
-            } catch (t: Throwable) {
-                logger.error("Error while locating target")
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/amplitude/android/internal/ViewHierarchyScanner.kt
```diff
@@ -72,21 +72,24 @@ internal object ViewHierarchyScanner {
...
+                                return@any true
+                            } else {
+                                return newTarget
+                            }
                         }
+                        false
                     }
-                    false
                 }
+            } catch (t: Throwable) {
```
polbins, line 91: Can we add the specific throwable that we're looking for here? Or if there's no specific throwable, maybe we should log the error when target ends up being null after trying to locate it.

android/src/main/java/com/amplitude/android/internal/locators/ComposeViewTargetLocator.java
```diff
@@ -72,27 +74,44 @@ public ViewTarget locate(
...
-                                semanticsModifierCore.getSemanticsConfiguration();
-                        for (Map.Entry<? extends SemanticsPropertyKey<?>, ?> entry : semanticsConfiguration) {
-                            final @Nullable String key = entry.getKey().getName();
-                            if ("OnClick".equals(key)) {
-                                isClickable = true;
-                            } else if ("TestTag".equals(key)) {
-                                if (entry.getValue() instanceof String) {
-                                    lastKnownTag = (String) entry.getValue();
+                    final Modifier modifier = modifierInfo.getModifier();
+                    if (modifier instanceof InspectableValue) {
```
polbins, line 78: the PR mentions that compose update changed from `SemanticsModifier` to `TestTagElement`, so does it mean we should still keep that old implementation for backwards compatibility?

android/src/main/java/com/amplitude/android/internal/locators/ComposeViewTargetLocator.java
```diff
@@ -72,27 +74,44 @@ public ViewTarget locate(
...
+                        final InspectableValue inspectableValue = (InspectableValue) modifier;
```
polbins, line 79: this might be a good time to convert this to Kotlin too, to get inferred types make it more succinct

igor-amp: no, bcos this is a java hack for reflection

android/src/main/java/com/amplitude/android/internal/locators/ComposeViewTargetLocator.java
igor-amp: in description I highlighted that this solution is backwards compatible with previous versions bcos it uses a different approach for gathering data.
We only left old click handling bcos it's a completely different handling in compose internal.

android/src/main/java/com/amplitude/android/internal/locators/ComposeViewTargetLocator.java
polbins: I'm not sure I understand, it seems we don't use reflection here specifically?

android/src/main/java/com/amplitude/android/internal/locators/ComposeViewTargetLocator.java
polbins: Makes sense, thanks for pointing that out!

polbins (approved):
I haven't tested this manually, but changes LGTM!
I hope the internal implementation for Compose sticks with this approach 🤞🏽

PR #262, opened 2025-04-03
fix: 4xx-handling-when-on-proxy

polbins:
### Summary

We now treat 400 errors differently based on responses that might be returned when connected to a proxy. In this PR we opted to use the `eventsToDrop` as an indicator if the response is coming from Amplitude BE (meaning it includes the required invalid fields), or if it's from a proxy.

### Checklist


igor-amp (approved):
lgtm

PR #263, opened 2025-04-09
refactor/test: remove reflection on unit tests

polbins:
### Checklist


core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -41,7 +41,7 @@ import java.util.concurrent.Executors
  * This is the SDK instance class that contains all of the SDK functionality.<br><br>
  * Many of the SDK functions return the SDK instance back, allowing you to chain multiple methods calls together.
  */
-open class Amplitude internal constructor(
+open class Amplitude(
```
polbins, line 44: I had to open this up for testability

core/src/main/java/com/amplitude/core/utilities/InMemoryResponseHandler.kt
```diff
@@ -76,9 +76,13 @@ internal class InMemoryResponseHandler(
...
-        eventPipeline.flushSizeDivider.incrementAndGet()
-        eventsList.forEach {
-            eventPipeline.put(it)
+
+        scope.launch(storageDispatcher) {
+            eventPipeline.flushSizeDivider.incrementAndGet()
+            eventsList.forEach {
+                delay(BACK_OFF)
+                eventPipeline.put(it)
+            }
```
polbins, line 85: I don't like to add this, but the current response handler integration tests are made in such a way that requires adding delays to control the flow of the test. This requires a lot more effort to fix, so it might be good to revisit when we're dealing with `InMemoryStorage` tasks.

core/src/main/java/com/amplitude/core/utilities/InMemoryResponseHandler.kt
```diff
@@ -43,7 +43,7 @@ internal class InMemoryResponseHandler(
         eventsString: String,
     ): Boolean {
         val eventsList = events as List<BaseEvent>
-        if (eventsList.size == 1 || badRequestResponse.isInvalidApiKeyResponse()) {
```
igor-amp, line 46: this change is intentional?

core/src/test/kotlin/com/amplitude/core/utils/FakeAmplitude.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.amplitude.core.State
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.TestDispatcher
+import kotlinx.coroutines.test.TestScope
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class FakeAmplitude(
```
igor-amp, line 14: do we really need fake? Can't we use spy?

android/src/test/kotlin/com/amplitude/android/AmplitudeTest.kt
```diff
@@ -0,0 +1,308 @@
...
+            scheduler = testScheduler,
+            configuration = createConfiguration()
+        )
+        amplitude.isBuilt.await()
+        amplitude.setUserId("test user")
+        amplitude.setDeviceId("test device")
+        advanceUntilIdle()
+        assertEquals("test user", amplitude.store.userId)
+        assertEquals("test device", amplitude.store.deviceId)
+        assertEquals("test user", amplitude.getUserId())
```
igor-amp, line 100: 2 times check to be 200% sure?

core/src/main/java/com/amplitude/core/utilities/InMemoryResponseHandler.kt
polbins: Yes, it's tied to previous changes here: https://github.com/amplitude/Amplitude-Kotlin/pull/262 
I added a test `test handle bad request response retry` that covers this case, and it required this change from the `InMemoryResponseHandler` which is used by the test.

android/src/test/kotlin/com/amplitude/android/AmplitudeTest.kt
```diff
@@ -0,0 +1,308 @@
...
+        amplitude.isBuilt.await()
+        assertNotNull(amplitude.store.deviceId)
+        assertNotNull(amplitude.getDeviceId())
+    }
+
+    @Test
+    fun amplitude_should_set_deviceId_from_configuration() = runTest {
+        val testDeviceId = "test device id"
+        // set device Id in the config
+        amplitude = Amplitude(createConfiguration(deviceId = testDeviceId))
```
igor-amp, line 209: it looks like we have `amplitude` variable set up in test class but never used.
Suggestion: move amplitude creation into @Before setup function.

android/src/test/kotlin/com/amplitude/android/AmplitudeTest.kt
```diff
@@ -0,0 +1,308 @@
...
+
+        amplitude.isBuilt.await()
+        val event = BaseEvent()
+        event.eventType = "test event"
+        amplitude.track(event)
+        advanceUntilIdle()
+        Thread.sleep(100)
+
+        val track = slot<BaseEvent>()
+        verify { mockedPlugin.track(capture(track)) }
```
igor-amp, line 128: StubPlugin is your own plugin, there is no need to use slot for it. Just add a list of event inside it and work with it directly.

igor-amp (approved):

PR #265, opened 2025-04-15
feat: add network tracking plugin

polbins:
### Summary

Adds `NetworkTracking` feature through the `NetworkTrackingPlugin`, only supports OkHttp for now.
- adds a `compileOnly` dependency to OkHttp

### Checklist


core/src/main/java/com/amplitude/core/network/AmplitudeNetworkTrackingPlugin.kt
```diff
@@ -0,0 +1,116 @@
...
+import okhttp3.Interceptor.Chain
+import okhttp3.Request
+import okhttp3.Response
+import okio.IOException
+
+private const val STAR_WILDCARD = "*"
+
+class AmplitudeNetworkTrackingPlugin(
+    private val captureRules: List<AmplitudeCaptureRule> = listOf(
+        AmplitudeCaptureRule(
```
sojingle, line 30: The `Amplitude` prefix can be removed from `AmplitudeCaptureRule` and `AmplitudeNetworkTrackingPlugin` because namespace is enough to identify.

polbins: I love that idea, let me make that change :)

core/src/main/java/com/amplitude/core/network/NetworkTrackingOptions.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    private val ignoreHostMatcher = HostMatcher(ignoreHosts)
+
+    fun shouldIgnore(host: String): Boolean {
+        return ignoreHostMatcher.matches(host)
+    }
+}
+
+private class HostMatcher(hosts: List<String>) {
+    private val hostRegexes: List<Regex> = hosts
```
igor-amp, line 53: can we also use lazy here as we did above?

core/src/main/java/com/amplitude/core/network/NetworkTrackingPlugin.kt
```diff
@@ -0,0 +1,140 @@
...
+            throw error
+        }
+    }
+
+    private fun shouldCapture(
+        request: Request,
+        responseCode: Int,
+    ): Boolean = with(options) {
+        val host = request.url.host
+        return when {
```
igor-amp, line 66: optional nit: can be simplified
`val host = request.url.host
return !shouldIgnore(host) &&
       !(ignoreAmplitudeRequests && request.amplitudeApiRequest()) &&
       (captureRules.matches(host, responseCode) || request.amplitudeApiRequest())`

core/src/main/java/com/amplitude/core/network/NetworkTrackingPlugin.kt
```diff
@@ -0,0 +1,140 @@
...
+            else -> false
+        }
+    }
+
+    /**
+     * Checks if the request is an Amplitude request, and not a [NETWORK_TRACKING] event as we don't
+     * want to track our own network tracking events to be tracked again for cases where the same
+     * [okhttp3.OkHttpClient] instance is used for both Amplitude and other network requests.
+     */
+    private fun Request.amplitudeApiRequest(): Boolean {
```
igor-amp, line 79: this one is great 👍

android/src/test/kotlin/com/amplitude/core/network/NetworkTrackingPluginTest.kt
```diff
@@ -0,0 +1,590 @@
...
+                    assertEquals("https://example.com/test", eventProperties[NETWORK_TRACKING_URL])
+                    assertEquals(200, eventProperties[NETWORK_TRACKING_STATUS_CODE])
+                    assertEquals(true, eventProperties[NETWORK_TRACKING_DURATION] != null)
+                }
+            )
+        }
+    }
+
+    @Test
+    fun `successful response outside capture range - does not trigger trackEvent`() {
```
igor-amp, line 81: No, a 404 response is not considered a successful response. In HTTP status codes, a successful response typically falls within the range of 200 to 299. A 404 indicates "Not Found", which is a client error and falls within the 400 range of status codes.

igor-amp (approved):

android/src/test/kotlin/com/amplitude/core/network/NetworkTrackingPluginTest.kt
polbins: I agree that 404 is not a successful response. The test is about a Not OK test case `- does not trigger trackEvent` which is what it does here.

igor-amp: I think it's just a bad test name in this case.

PR #267, opened 2025-04-17
fix: update network listener to cover cases where network is blocked

polbins:
### Summary

When the app is backgrounded, the network actually gets blocked (`onBlockedStatusChanged(...)`). We want to set our app to offline when this happens.
As part of this PR, we've also shortened the feedback loop once the internet goes back through `onCapabilitiesChanged(...)` callback. 

### Checklist


igor-amp (approved):
lgtm
nit: It feels like we can unite NetworkCallback and NetworkState into single class.

PR #276, opened 2025-05-22
fix: use range for common success/failure status codes

polbins:
### Summary

Support for common success/failure status codes in AnalyticsResponse.

### Checklist

### DIFF
--- a/android/src/test/kotlin/com/amplitude/android/ResponseHandlerIntegrationTest.kt
+++ b/android/src/test/kotlin/com/amplitude/android/ResponseHandlerIntegrationTest.kt
@@ -114,7 +114,7 @@ class ResponseHandlerIntegrationTest {
         amplitude.isBuilt.await()
 
         // verify it will attempt to retry, but single event file will be removed
-        val code = HttpStatus.PAYLOAD_TOO_LARGE.code
+        val code = HttpStatus.PAYLOAD_TOO_LARGE.range.first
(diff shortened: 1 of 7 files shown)


android/src/test/kotlin/com/amplitude/android/ResponseHandlerIntegrationTest.kt
```diff
@@ -114,7 +114,7 @@ class ResponseHandlerIntegrationTest {
         amplitude.isBuilt.await()
 
         // verify it will attempt to retry, but single event file will be removed
-        val code = HttpStatus.PAYLOAD_TOO_LARGE.code
+        val code = HttpStatus.PAYLOAD_TOO_LARGE.range.first
```
igor-amp, line 117: If for some reason HttpStatus.PAYLOAD_TOO_LARGE.range is ever an empty range (e.g., IntRange.EMPTY or similar), .first would throw an exception.

igor-amp (approved):
Please don't use `Collection.first`, it's almost alway `firstOrNull`.

polbins:
> Please don't use `Collection.first`, it's almost alway `firstOrNull`.

`HttpStatus` is designed such that it will never have an empty range (see the constructor), so this concern should not occur in practice. I could make an extension function for the single value ones, which guarantees non-null value. I think that would be preferred instead of `firstOrNull` because then the code paths will need to handle the null case.

PR #277, opened 2025-05-22
Fix amplitude host detection in network tracking plugin

ppatibandla:
## Summary
- fix detection logic for Amplitude API requests in `NetworkTrackingPlugin`

## Testing
- `./gradlew test` *(fails: Downloading gradle distribution requires network access)*


ppatibandla:
PR created by OpenAI codex agent. @igor-amp @polbins please take a look to see if it makes sense or not.

polbins:
Saw this just now @ppatibandla, I think it tries to be more explicit for the `body == null` case. But no functional change so I think we should just keep it as-is.

PR #279, opened 2025-05-23
Move event bridge module into core

igor-amp:
## Summary
- move `EventBridge` and `EventBridgeContainer` sources to `core`
- remove `event-bridge` module from build
- update Android and core gradle dependencies

## Testing
- `./gradlew test` *(fails: No route to host)*


polbins (approved):
Nice! It's great that it can be used for things like this :D

PR #286, opened 2025-05-30
fix: deprecate on enter / exit foreground

polbins:
### Summary

Upon testing, this exposed function can produce these end/start session events if used incorrectly.
This PR is removing the implementation from Amplitude and moving it directly into Timeline.

### Checklist


igor-amp (approved):
this is wild to check. lgtm

PR #293, opened 2025-06-06
Move module deps to version catalog

igor-amp:
## Summary
- centralize Android and core module dependencies in libs.versions.toml
- reference the catalog from module build files

## Testing
- `./gradlew test` *(fails: Unable to tunnel through proxy)*

------
https://chatgpt.com/codex/tasks/task_b_684374adc4308331aabac348b8f0cdd5

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,18 +1,56 @@
 [versions]
+activityKtx = "1.9.1"
+analyticsConnector = "1.0.0"
+composeUi = "1.6.8"
+coreKtx = "1.7.0"
 coroutines = "1.5.2"
 coroutinesTest = "1.6.4"
+espressoCore = "3.4.0"
+fragmentKtx = "1.8.2"
 json = "20211205"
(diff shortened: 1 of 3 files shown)


gradle/libs.versions.toml
```diff
@@ -1,18 +1,56 @@
...
 coroutines = "1.5.2"
 coroutinesTest = "1.6.4"
+espressoCore = "3.4.0"
+fragmentKtx = "1.8.2"
 json = "20211205"
 junit = "5.7.2"
-mockk = "1.12.3"
+junit4 = "4.13.2"
+kotlin = "1.8.0"
+mockito = "4.3.1"
```
polbins, line 14: This is concerning, we have mockk and mockito. Maybe we should choose one soon and deprecate the other, before integrating to SR

polbins (approved):

gradle/libs.versions.toml
igor-amp: yeap, let's remove one

Issue #316, opened 2025-07-14
ScreenView tracking does not work with jetpack fragment navigation


ajaugust44:
## SDK Version and configuration details
1.21.4
No plugins
Configured with autoconfigureOptions `appLifecycles` and `screenViews`

## Expected Behavior
[Navigating to a fragment](https://developer.android.com/guide/navigation/design#fragments) using the jetpack navigation framework with screenViews enabled should fire a Fragment Viewed event

## Current Behavior
Navigating to a fragment using a support fragment manager transaction causes Fragment Viewed events, but using an XML nav graph or a kotlin-dsl nav graph as [recommended by the android docs](https://developer.android.com/guide/navigation/design#fragments) does not trigger any event.

## Possible Solution
I _think_ that updating [this line](https://github.com/amplitude/Amplitude-Kotlin/blob/2a6d8aeaa487b71503bc2c1d7ba9b3fc698ea501/android/src/main/java/com/amplitude/android/internal/fragments/FragmentActivityHandler.kt#L20C13-L20C87) to use `recursive = true` may solve this

## Steps to Reproduce
 

1. Build the attached test project (add your api key in a local.properties file), or equivalent with jetpack fragment nav set up
2. Keep your eye on the logcat or the text view with logs
3. Tap into the kotlin dsl and nav graph resource fragments
4. Note that the log does not update
5. Tap into the transaction fragment
6. Note that the log does update

[Test project](https://github.com/user-attachments/files/21223493/AmplitudeFragmentTracking.zip)

Issue #315, opened 2025-07-14
AutocaptureOptions must contain +appLifecycles for screenView tracking to work


ajaugust44:
## SDK Version and configuration details

1.21.4

No plugins
  
autocapture +screenViews

## Expected Behavior
Following the sdk [documentation](https://amplitude.com/docs/sdks/analytics/android/android-kotlin-sdk#track-screen-views) to add screen view tracking- 
```
Configuration(
  ...,
  autocaptureOptions { +screenViews }
)
```
should enable screen view tracking

## Current Behavior
The autocaptureOptions must also include `+appLifecycles` to enable screen view tracking

## Possible Solution
Update the [documentation](https://amplitude.com/docs/sdks/analytics/android/android-kotlin-sdk#track-screen-views) for screen view tracking to include the `+appLifecycles` requirement 
OR
Add an `|| SCREEN_VIEWS in configuration.autocapture` to [this line](https://github.com/amplitude/Amplitude-Kotlin/blob/2a6d8aeaa487b71503bc2c1d7ba9b3fc698ea501/android/src/main/java/com/amplitude/android/Amplitude.kt#L63) 
I'm not sure whether there would be unwelcome side effects/new bugs from the latter

## Steps to Reproduce
1. Set configuration autoCapture options to only include +screenViews
2. Open fragments and note that no fragment views are tracked
3. Update configuration to include +appLifecycles
4. Open fragments and note that fragment views are tracked

PR #317, opened 2025-07-15
fix: remove coupling of autocapture options to lifecycle

polbins:

Current implementation requires `appLifecycles` autocapture to be enabled so that `screenViews` will work. see https://github.com/amplitude/Amplitude-Kotlin/issues/315

There's also a nav graph issue (https://github.com/amplitude/Amplitude-Kotlin/issues/316) reported and a possible fix for it is shipped as part of this.

### Describe the solution

Check autocapture options if it requires the activity lifecycle callbacks, instead of checking `appLifecycles` alone. This will remove coupling in between autocapture options.

Enable `recursive` for `registerFragmentLifecycleCallbacks` to register to child FragmentManagers as well.

Manual and Unit tests 

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/AutocaptureOptions.kt
+++ b/android/src/main/java/com/amplitude/android/AutocaptureOptions.kt
@@ -20,7 +20,7 @@ enum class AutocaptureOption {
     DEEP_LINKS,
 
     /**
-     * Enable screen view tracking.
+     * Enable screen view tracking for Activities and Fragments (including nested ones).
      */
     SCREEN_VIEWS,
 


android/src/main/java/com/amplitude/android/internal/fragments/FragmentActivityHandler.kt
```diff
@@ -17,7 +17,7 @@ internal object FragmentActivityHandler {
     ) {
         (this as? FragmentActivity)?.apply {
             val callback = AutocaptureFragmentLifecycleCallbacks(track, logger)
-            supportFragmentManager.registerFragmentLifecycleCallbacks(callback, false)
+            supportFragmentManager.registerFragmentLifecycleCallbacks(callback, true)
```
igor-amp, line 20: I think it's acceptable, but Let's add comment to SCREEN_VIEWS enum that we track ALL fragments recursively if this is enabled.

igor-amp (approved):

PR #327, opened 2025-08-21
refactor: remote config - part 1

polbins:

Implementation for remote config on Kotlin SDK, please see [design doc](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/3255205972/Design+Doc+Kotlin+Android+Remote+Config+to+Core+migration) part 1.

### Describe the solution

- extend `HttpClient` to support generic requests
- include new keys to `Storage` to support remote config
- add `RemoteConfigClient` which manages querying and subscribing to Remote config changes
  - (still lacking some features, will be completed in part 2)
- add essential tests to `HttpClient` and `RemoteConfigClient`

- unit tests
- manual tests using the sample app 

[Design Doc](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/3255205972/Design+Doc+Kotlin+Android+Remote+Config+to+Core+migration)

### Checklist

### DIFF
--- a/core/src/test/kotlin/com/amplitude/core/remoteconfig/RemoteConfigClientTest.kt
+++ b/core/src/test/kotlin/com/amplitude/core/remoteconfig/RemoteConfigClientTest.kt
@@ -207,9 +207,6 @@ class RemoteConfigClientTest {
             assertTrue(tempCallbackInvocations >= 2)
             // Persistent callback should have received multiple remote updates across steps
             assertTrue(persistentCallbackInvocations >= 3)
-
-            // Whether GC cleanup happened or not, the system continues to function
-            assertTrue(true, "Weak reference subscription system handles temp callbacks correctly")
         }
 
     @Test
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -0,0 +1,489 @@
...
+    override fun subscribe(
+        key: Key,
+        callback: RemoteConfigClient.RemoteConfigCallback,
+    ) {
+        // Clean up dead weak references before adding new one
+        cleanupDeadReferences()
+
+        // Add new weak reference callback
+        val subscriberList =
+            keySpecificSubscribers[key.value]
```
igor-amp, line 126: nit: getOrPut

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -0,0 +1,489 @@
...
+                configKey = key.value,
+                index = subscriberList.indexOf(weakCallback),
+                config = storedData.config,
+                source = CACHE,
+                timestamp = storedData.timestamp,
+            )
+        }
+
+        // Trigger remote fetch to ensure latest config is available
+        updateConfigs()
```
igor-amp, line 147: It looks like right now if we subscribe for different keys or for the same key from different plugins you are going to call `updateConfigs()` on each of them? Since it runs on networkIODispatcher and I don't think we cache network, will it result in many network calls?

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -0,0 +1,489 @@
...
+    private fun getAllStoredConfigs(): Map<String, ConfigMap> {
+        return try {
+            val configJson = storage.read(REMOTE_CONFIG)
+            if (configJson.isNullOrBlank()) {
+                logger.debug("No stored config found in storage")
+                return emptyMap()
+            }
+
+            val allConfigs = JSONObject(configJson).toMapObj()
+            val result = mutableMapOf<String, ConfigMap>()
```
igor-amp, line 213: nit: buildMap

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -0,0 +1,489 @@
...
+    private fun notifySubscribersForKey(
+        configKey: String,
+        index: Int? = null,
+        config: ConfigMap,
+        source: RemoteConfigClient.Source,
+        timestamp: Long,
+    ) {
+        val subscriberList = keySpecificSubscribers[configKey] ?: return
+        val notifySubscribers =
+            index?.let { subscriberList.getOrNull(it) }
```
igor-amp, line 304: nit: filterIndexed

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -0,0 +1,489 @@
...
+        // Clean up dead weak references before adding new one
+        cleanupDeadReferences()
+
+        // Add new weak reference callback
+        val subscriberList =
+            keySpecificSubscribers[key.value]
+                ?: CopyOnWriteArrayList<WeakCallback>()
+        val weakCallback = WeakCallback(WeakReference(callback))
+        subscriberList.add(weakCallback)
+        keySpecificSubscribers[key.value] = subscriberList
```
igor-amp, line 130: hm, after making a revision of code below, I have a question. subscribe is not single threaded. It looks like there is a case where 2 concurrent subscriptions can override `keySpecificSubscribers[key.value] = subscriberList`. I also don't really see the point in adding and removing these keys on subscribe and cleanup.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
igor-amp: SR is a single feature and config refreshes are connected to some specific events: bootstrap, sessionId change or other that should not happen frequently and "in theory" should not overlap. Here we will be calling it many times in short period of time. Network IO dispatcher is indeed a single thread but you still enqueue many coroutines to it that do the same job again.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -0,0 +1,489 @@
...
+        timestamp: Long,
+    ) {
+        val subscriberList = keySpecificSubscribers[configKey] ?: return
+        val notifySubscribers =
+            index?.let { subscriberList.getOrNull(it) }
+                ?.let { listOf(it) }
+                ?: subscriberList
+        notifySubscribers.forEach { weakCallback ->
+            try {
+                weakCallback.get()?.run {
```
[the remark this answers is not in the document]
igor-amp: Damn, I hate to be sick. What I mean: your WeakCallback class is effectively just a wrappper around WeakRef. Don't propagate weak ref outside of it, just use lambda to propagate the action on the object itself. Classes that use your wrapper don't need to know how it's implemented inside.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -0,0 +1,533 @@
...
+        // Clean up dead weak references before adding new one
+        cleanupDeadReferences()
+
+        // Add new weak reference callback
+        val subscriberList =
+            keySpecificSubscribers.getOrPut(key.value) {
+                CopyOnWriteArrayList<WeakCallback>()
+            }
+        val weakCallback = WeakCallback(callback)
+        subscriberList.add(weakCallback)
```
igor-amp, line 144: I wonder if java WeakHashMap is solving the same problem as you solve here manually.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -0,0 +1,533 @@
...
+        private val weakRef = WeakReference(callback)
+
+        fun isAlive(): Boolean = weakRef.get() != null
+
+        fun runSafely(action: RemoteConfigClient.RemoteConfigCallback.() -> Unit) {
+            val callback = weakRef.get() ?: return
+            try {
+                callback.action()
+            } catch (e: Exception) {
+                logger.error("Exception in subscriber callback: ${e.message}")
```
igor-amp, line 112: We reference parent logger object here and keep parent as a reference. But it looks like it should not leak.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -0,0 +1,533 @@
...
+        index: Int? = null,
+        config: ConfigMap,
+        source: RemoteConfigClient.Source,
+        timestamp: Long,
+    ) {
+        val subscriberList = keySpecificSubscribers[configKey] ?: return
+        val notifySubscribers =
+            subscriberList
+                .filterIndexed { i, _ -> index == null || i == index }
+                .ifEmpty { subscriberList }
```
igor-amp, line 351: if we fail to notify a specific subscriber we notify all instead? why this logic?

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -0,0 +1,533 @@
...
+            weakCallback.runSafely {
+                onUpdate(config, source, timestamp)
+            }
+        }
+    }
+
+    /**
+     * Clean up dead weak references to prevent memory accumulation.
+     */
+    private fun cleanupDeadReferences() {
```
igor-amp, line 362: same question about WeakHashMap. We intentionally remove wrappers for GC to collect them when GC can just collect them.

core/src/test/kotlin/com/amplitude/core/remoteconfig/RemoteConfigClientTest.kt
```diff
@@ -0,0 +1,1217 @@
...
+            testDispatcher.scheduler.advanceUntilIdle()
+
+            // The test demonstrates that weak reference system works
+            // Counter shows temp callback were invoked initially and then cleaned up
+            assertTrue(tempCallbackInvocations >= 2)
+            // Persistent callback should have received multiple remote updates across steps
+            assertTrue(persistentCallbackInvocations >= 3)
+
+            // Whether GC cleanup happened or not, the system continues to function
+            assertTrue(true, "Weak reference subscription system handles temp callbacks correctly")
```
igor-amp, line 212: This line is not a test

PR #337, opened 2025-12-18
fix: resolve race condition in AndroidLifecyclePlugin causing duplicate session_start events

AntonnyAGS:
## Problem
AndroidLifecyclePlugin had a race condition that caused duplicate session_start events when activities were recreated or when the app returned from background. This was especially problematic with deep links autocapture enabled, creating "ghost sessions" with only session_start events.

## Root Cause
The plugin checked `started.isEmpty()` BEFORE adding the activity to the set, causing `onEnterForeground()` to be triggered incorrectly when the same activity returned from background.

## Solution
- Moved `started.add(activity.hashCode())` to execute BEFORE the foreground check
- Changed condition from `started.isEmpty()` to `started.size == 1` after adding
- This ensures atomic state update following the check-then-act pattern

## Changes
- AndroidLifecyclePlugin.kt: Fixed race condition in onActivityStarted method
- AndroidLifecyclePluginTest.kt: Added regression test for multi-activity scenarios

## Testing
- All existing tests pass (169 tests)
- New test verifies onEnterForeground is only called during true foreground transitions
- Prevents duplicate session creation in multi-activity and deep link scenarios

### Checklist

---

> [!NOTE]
> Ensure foreground enter/open events trigger only on first activity start and exit on last stop; add test covering multi-activity lifecycle transitions.
> 
> - **Android**:
>   - `android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt`:
>     - In `onActivityStarted`, add activity to `started` before checks and use `wasEmpty` to gate `onEnterForeground` and `APPLICATION_OPENED` events.
>     - Retains deep link and screen viewed tracking on start.
> - **Tests**:
>   - `android/src/test/kotlin/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt`:
>     - Add test ensuring `onEnterForeground` fires only once across multiple started activities and `onExitForeground` triggers only when the last activity stops.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c7f85c46039cb9d5f28a634aa7e6c2689b923546. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
+++ b/android/src/main/java/com/amplitude/android/plugins/AndroidLifecyclePlugin.kt
@@ -124,13 +124,14 @@ class AndroidLifecyclePlugin(
             onActivityCreated(activity, activity.intent.extras)
         }
 
-        if (started.isEmpty()) {
+        val wasEmpty = started.isEmpty()
+        started.add(activity.hashCode())
+
+        if (wasEmpty) {
             androidAmplitude.onEnterForeground(System.currentTimeMillis())
         }
 
-        started.add(activity.hashCode())
-
-        if (autocaptureState.appLifecycles && started.size == 1) {
+        if (autocaptureState.appLifecycles && wasEmpty) {
             DefaultEventUtils(androidAmplitude).trackAppOpenedEvent(
                 packageInfo = packageInfo,
                 isFromBackground = appInBackground,
--- a/android/src/test/kotlin/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt
+++ b/android/src/test/kotlin/com/amplitude/android/plugins/AndroidLifecyclePluginTest.kt
@@ -391,6 +391,55 @@ class AndroidLifecyclePluginTest {
             close()
         }
 
+    @OptIn(com.amplitude.android.GuardedAmplitudeFeature::class)
+    @Test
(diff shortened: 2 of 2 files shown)


polbins:
Hi @AntonnyAGS, can you [file an issue](https://github.com/amplitude/Amplitude-Kotlin/issues/new?template=Bug_report.md) please?  

And if you have logs showing the duplicate `session_start` events, and the exact navigation pattern that triggered this (e.g., deep link while app in foreground, singleTask activity, multi-window mode, etc.) that would be great. Thanks!

polbins:
> Hi @polbins! We discovered this issue in our production environment after enabling deep links in autocapture. We started seeing "ghost sessions" that contained ONLY session_start events with no other activity, which was causing incorrect analytics reports. After investigating, we found the race condition in AndroidLifecyclePlugin where onEnterForeground() was being triggered incorrectly during multi-activity transitions (like when handling deep links while the app was already in the foreground). We created tests to reproduce and verify the fix - the test in `AndroidLifecyclePluginTest.kt:394` simulates the scenario that was creating these incomplete sessions. Should I file a separate issue with the production scenario details and link it to this PR?

Yes please, the bug report template contains the details that we need, and please free to add some more info that are relevant.

PR #349, opened 2026-02-02
fix: queue user and device ID changes for proper event ordering

polbins:
## Describe what this PR is addressing

Identity changes `setUserId()` / `setDeviceId()` / `reset()` followed by `track()` could send events with stale identity because identity changes and events were processed on different dispatchers with no ordering guarantee.

  - Identity changes ran on `amplitudeDispatcher`
  - Events processed on `storageIODispatcher`

  > No ordering guarantee between the two.

## Describe the solution

Dual-path architecture:
 1. **Immediate** - `store.userId = value` updates in-memory state synchronously for event enrichment and plugin notifications
2. **Background** - `super.setUserId()` dispatches `idContainer.commit()` via coroutine for disk persistence

Key changes:
  - `Amplitude.kt`: Android overrides update `store` immediately then delegate to core for persistence
  - `State.kt`: Unified `Identity` object with synchronized access and idempotent observer notifications
  - `AndroidContextPlugin.kt`: Refactored to pure `createDeviceId()` function

## Steps to verify the change

Manual verification:
- Call `setUserId("user1")` then immediately `track("event1")`
  - Verify event1 has userId "user1"
- Call `reset()` then immediately `track("event2")`
  - Verify event2 has a new deviceId (not the previous one)

## Checklist
- [x] Does your PR title have the correct title format?
- [ ] Does your PR have a breaking change? **No**

---

> [!NOTE]
> **Medium Risk**
> Touches core identity/state synchronization and device ID initialization paths, which can affect event attribution and persistence across platforms despite added test coverage.
> 
> **Overview**
> Fixes cases where `setUserId()`/`setDeviceId()`/`reset()` followed by `track()` could emit events with stale identity by **updating in-memory identity (`State`) synchronously** and only committing to the identity manager asynchronously.
> 
> Android initialization now **computes and persists `deviceId` during `buildInternal`** (preferring configured ID, then stored non-AppSet IDs, then generated), while `AndroidContextPlugin` is refactored to a pure `createDeviceId()` helper and deprecates plugin-driven identity setting. Core `State` is reworked to hold a synchronized `Identity` object with idempotent observer notifications, `ContextPlugin` no longer performs deviceId initialization, and new tests assert ordering for `setUserId`, `setDeviceId`, `identify` options, and `reset()`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4203557bf39407e91ac30283dd1d784312a540d8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -22,6 +22,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.launch
 import java.io.File
 import java.util.concurrent.Executors
 import com.amplitude.core.Amplitude as CoreAmplitude
@@ -103,7 +104,9 @@ open class Amplitude internal constructor(
             configuration.deviceId
                 ?: store.deviceId?.takeIf { validDeviceId(it, allowAppSetId = false) }
                 ?: androidContextPlugin.createDeviceId()
-        setDeviceId(deviceId)
+        // Persist device ID synchronously during build (not via coroutine dispatch)
+        store.deviceId = deviceId
+        idContainer.identityManager.editIdentity().setDeviceId(deviceId).commit()
         add(GetAmpliExtrasPlugin())
         add(AndroidLifecyclePlugin(activityLifecycleCallbacks))
         add(AnalyticsConnectorIdentityPlugin())
@@ -130,23 +133,20 @@ open class Amplitude internal constructor(
      * @return the Amplitude instance
      */
     override fun reset(): Amplitude {
-        if (!isBuilt.isCompleted) {
-            logger.error("Cannot reset identity before Amplitude is initialized.")
-            return this
+        amplitudeScope.launch(amplitudeDispatcher) {
+            isBuilt.await()
+            setUserId(null)
+            setDeviceId(androidContextPlugin.createDeviceId())
         }
-        setUserId(null)
-        setDeviceId(androidContextPlugin.createDeviceId())
         return this
     }
 
     override fun setUserId(userId: String?): Amplitude {
-        store.userId = userId
         super.setUserId(userId)
         return this
     }
 
     override fun setDeviceId(deviceId: String): Amplitude {
-        store.deviceId = deviceId
         super.setDeviceId(deviceId)
         return this
     }
--- a/core/src/main/java/com/amplitude/core/Amplitude.kt
+++ b/core/src/main/java/com/amplitude/core/Amplitude.kt
@@ -300,6 +300,7 @@ open class Amplitude(
      * @return the Amplitude instance
      */
     open fun setUserId(userId: String?): Amplitude {
+        store.userId = userId
         amplitudeScope.launch(amplitudeDispatcher) {
             if (isBuilt.await()) {
                 idContainer.identityManager.editIdentity().setUserId(userId).commit()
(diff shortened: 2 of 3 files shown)


android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -113,17 +113,31 @@ open class Amplitude internal constructor(
...
-        this.setUserId(null)
-        amplitudeScope.launch(amplitudeDispatcher) {
-            isBuilt.await()
-            idContainer.identityManager.editIdentity().setDeviceId(null).commit()
-            androidContextPlugin.initializeDeviceId(configuration as Configuration)
+        if (!isBuilt.isCompleted) {
+            logger.error("Cannot reset identity before Amplitude is initialized.")
+            return this
         }
+        val newDeviceId = configuration.deviceId ?: androidContextPlugin.createDeviceId()
```
crleona: Reset probably shouldn't reuse the same `configuration.deviceId` that was set initially? does this match the previous logic?

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -294,27 +288,17 @@ open class Amplitude(
...
-
     /**
      * Sets a custom device id. <b>Note: only do this if you know what you are doing!</b>
      *
      * @param deviceId custom device id
      * @return the Amplitude instance
      */
-    fun setDeviceId(deviceId: String): Amplitude {
+    open fun setDeviceId(deviceId: String): Amplitude {
         amplitudeScope.launch(amplitudeDispatcher) {
```
crleona: do we need to dispatch these user/device id updates in a coroutine?

polbins: Yes, `:core` needs it for `isBuilt.await()` but `:android` override skips it.
I haven't really focused on `:core` that much, I'm not sure if it's being used in the wild.

android/src/main/java/com/amplitude/android/Timeline.kt
```diff
@@ -65,37 +71,48 @@ class Timeline(
...
+            is Event -> {
                 processEvent(message.event)
             }
-            is EventQueueMessage.ExitForeground -> {
+            is ExitForeground -> {
                 foreground.set(false)
                 refreshSessionTime(message.timestamp)
             }
+            is SetIdentity -> {
+                amplitude.idContainer.identityManager.editIdentity()
```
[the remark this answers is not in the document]
polbins: So we have a `store` (State.kt) which is for immediate updates. It notifies plugins including AnalyticsConnectorIdentityPlugin w/c bridges experiment SDK immediately.

And we have this `idContainer` which is for disk persistence across app restarts, and it's shared by both Analytics/Experiment. We're queueing it here together with the events so that persistence order is maintained. By processing them on the same queue, identity only persists to disk after all preceding events have been persisted.

android/src/main/java/com/amplitude/android/Amplitude.kt
crleona: huh, let's keep this behavior then. Double checked with Swift but we don't have a deviceId in configuration (yet)

android/src/main/java/com/amplitude/android/Timeline.kt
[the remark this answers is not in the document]
crleona: Is the timeline integration here to maintain expected ordering for device/user ids for events? I would prefer to set idContainer to save to disk and notify experiment as soon as possible, event processing is not always immediate.

This also represents a behavioral change for when device id is set on initial launch, previously:
1/ Amplitude initializes with no device id set
2/ AndroidContextPlugin generates / sets a new userId on idContainer immediately via the setDeviceIdInternal path -> this is now async on the event queue.
I think this is important to maintain because of how much they needed to override to get the behavior 😄.

On iOS, we rely on the same queue for background execution and event ordering / processing - in the original core amplitude design, was the background dispatch just to do the work in the background? If that is fairly immediate (ie, not waiting for event processing), it is probably ok to notify the plugins then.

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -126,17 +125,29 @@ open class Amplitude internal constructor(
...
      * @return the Amplitude instance
      */
     override fun reset(): Amplitude {
-        this.setUserId(null)
-        amplitudeScope.launch(amplitudeDispatcher) {
-            isBuilt.await()
-            idContainer.identityManager.editIdentity().setDeviceId(null).commit()
-            androidContextPlugin.initializeDeviceId(configuration as Configuration)
+        if (!isBuilt.isCompleted) {
+            logger.error("Cannot reset identity before Amplitude is initialized.")
```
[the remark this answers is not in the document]
polbins: It's been a couple of back and forths now, but the logger + early return is just for handling the edge case.

Made it use await instead.

PR #406, opened 2026-04-30
feat(core): Plugin contract for unified blades

polbins:
Blade SDKs (Session Replay, Experiment) need to observe identity/session/optOut
changes on their host Amplitude instance without subclassing `ObservePlugin` or
polling. Currently only `ObservePlugin` gets `onUserIdChanged`/`onDeviceIdChanged`,
and timeline plugins get nothing.

### Describe the solution
- Lifecycle callbacks (`onUserIdChanged`, `onDeviceIdChanged`, `onSessionIdChanged`,
  `onOptOutChanged`, `onReset`) added to `Plugin` with default no-ops — any plugin
  type can opt in.
- `ObservePlugin` overrides `onUserIdChanged`/`onDeviceIdChanged` as abstract
  (backward compat).
- `Plugin.name` for deduplication — same-name plugin replaces the old one (with
  teardown).
- `IdentityCoordinator` handles all identity mutations with notifications dispatched
  outside the synchronized block, eliminating the reentrancy / silent-clobber hazard.
- `State.setIdentity` batches field writes and notifies once.
- `Amplitude.add()` deduplicates by `Plugin.name` (teardown on evict).
- `findPlugin<T>()` searches both timeline and observe store using a snapshot, fixing
  the CME risk.
- `safelyNotify` isolates per-plugin throws in all fan-outs; fatal `Error`s propagate.
- `State.remove()` calls `teardown()`; teardown in dedup is wrapped in try-catch.

`./gradlew build apiCheck` — 30 plugin contract tests + full suite green.

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes identity mutation, reset semantics, and plugin notification ordering across core and Android; public API additions are opt-in restricted but behavior shifts affect all registered plugins.
> 
> **Overview**
> Extends the **plugin contract** so blade SDKs can react to identity, session, opt-out, and reset without subclassing `ObservePlugin` only. Any `Plugin` can implement optional callbacks (`onUserIdChanged`, `onDeviceIdChanged`, `onSessionIdChanged`, `onOptOutChanged`, `onReset`); **`Plugin.name`** enables **dedup on `add()`** (evict same name + `teardown`).
> 
> **Identity and reset behavior** is reworked: `IdentityCoordinator` commits user/device id **under lock** but notifies **after** releasing the lock (fixes reentrant plugin callbacks clobbering inner writes). **`resetIdentity`** / **`reset()`** emit **one bundled** identity notification plus **`onReset`**, not separate userId then deviceId fires. **`notifyAllPlugins`** fans out to timeline plugins and a **snapshotted** observe store, with **`safelyNotify`** so one throwing plugin does not break others.
> 
> **Timeline** gains an **Observe** mediator so `Plugin.Type.Observe` plugins are no longer dropped. **`findPlugin<T>()`** searches timeline and observe store. **Android** wires **`onSessionIdChanged`** when sessions start or restore from storage; **`reset()`** resolves device id via **`AndroidContextPlugin.resolveDeviceId`** before atomic reset.
> 
> Adds **core README**, **apiCheck**-visible APIs, and broad **contract/concurrency tests** (core + Android).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2102bb0cda8ad8c937b778f23f92308a11e1e05d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/core/src/main/java/com/amplitude/core/IdentityCoordinator.kt
+++ b/core/src/main/java/com/amplitude/core/IdentityCoordinator.kt
@@ -1,6 +1,7 @@
 package com.amplitude.core
 
 import com.amplitude.id.IdentityManager
+import java.util.EnumSet
 
 /**
  * Coordinates identity (userId/deviceId) updates between [State] and [IdentityManager].
@@ -8,6 +9,19 @@ import com.amplitude.id.IdentityManager
  * All mutations go through synchronized blocks to ensure atomic state + persistence.
  * Before build, only [State] is updated and pending flags are set.
  * During [bootstrap], persisted identity is reconciled with any pre-build changes.
+ *
+ * ## Reentrancy safety
+ *
+ * Plugin callbacks (e.g. [com.amplitude.core.platform.Plugin.onUserIdChanged]) can call
+ * back into [setUserId]/[setDeviceId] on the same thread. Because [synchronized] is
+ * reentrant on the JVM, a naive implementation that fires [State.onIdentityChanged] while
+ * holding the lock allows the inner call to fully commit, then the outer call's commit
+ * overwrites the inner value — a silent data-loss bug.
+ *
+ * Fix: under the lock we write to [State]'s backing fields silently (no notification) and
+ * commit to [IdentityManager]. After the lock is released we dispatch the notification.
+ * If a callback then calls [setUserId] again, that call acquires the lock, updates state,
+ * and dispatches its own notification — without any outer commit following it.
  */
 internal class IdentityCoordinator internal constructor(private val state: State) {
     private val lock = Any()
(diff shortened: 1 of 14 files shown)


polbins:
bugbot run

polbins:
bugbot run

polbins:
bugbot run

core/src/main/java/com/amplitude/core/PluginNotify.kt
```diff
@@ -0,0 +1,29 @@
...
+ * still get the isolation guarantee.
+ */
+internal fun safelyNotify(
+    plugin: Plugin,
+    logger: Logger?,
+    block: (Plugin) -> Unit,
+) {
+    try {
+        block(plugin)
+    } catch (e: Exception) {
```
sojingle, line 26: doc says to catch [throwable], but here catches Exception. Should change doc or catch Throwable instead

core/src/main/java/com/amplitude/core/IdentityCoordinator.kt
```diff
@@ -36,6 +36,25 @@ internal class IdentityCoordinator internal constructor(private val state: State
...
+            identityManager?.editIdentity()
+                ?.setUserId(null)
+                ?.setDeviceId(newDeviceId)
+                ?.commit()
+                ?: run {
+                    pendingUserId = Pending(null)
+                    pendingDeviceId = Pending(newDeviceId)
+                }
+        }
+    }
```
sojingle, line 80: setUserId / setDeviceId / resetIdentity all write State before they commit to IdentityManager, and the State setter fires `onUserIdChanged` / `onDeviceIdChanged` synchronously while still holding the coordinator's lock. Because synchronized is reentrant, a plugin that reacts to the callback by calling `amplitude.setUserId(...)` (or `setDeviceId`/`reset`) re-enters this class on the same thread and runs its own State+commit sequence — then control returns to the outer call, which commits its original value last and silently clobbers the nested write in persistence.

core/src/main/java/com/amplitude/core/PluginNotify.kt
polbins: Updated, thanks!

core/src/main/java/com/amplitude/core/IdentityCoordinator.kt
polbins: Fixed. All mutation paths (`setUserId`, `setDeviceId`, `resetIdentity`, `bootstrap`) now dispatch notifications outside the synchronized block. A re-entrant callback acquires the lock cleanly with no outer commit following it. Added a `## Reentrancy safety` section to the KDoc explaining the invariant.

polbins:
@cursor push 52d0f1e2e9

polbins:
closing this in favor of https://github.com/amplitude/Amplitude-Kotlin/pull/411

PR #409, opened 2026-05-05
feat(core): align RemoteConfigClient key model with iOS/Web

polbins:
Android pre-flattened the remote config blob and matched exact key strings. iOS and Web store the full nested blob and resolve a slice by dot-path at delivery time. This aligns Android's key model **and** delivery semantics with them.

### Describe the solution
- **Key model:** `Key` is now a `sealed class` — `data object`s (`AnalyticsSdk`, `Diagnostics`, `SessionReplayPrivacyConfig`, `SessionReplaySamplingConfig`) plus `Custom("dot.path")` for arbitrary sections. `Custom` is a plain class (no `copy()`/`componentN()` leak).
- **Fetch + resolve:** one nested blob via `config_group=android`; slices resolved by dot-path at delivery (no pre-flattening).
- **Callback contract:** `onUpdate(config: ConfigMap?, source, timestamp: Long?)` — `timestamp == null` ⇒ failed fetch; non-null timestamp + `config == null` ⇒ key absent.
- **Delivery:** the mode governs only the **initial** callback; afterward every subscriber receives subsequent successful fetches as updates. `All` = cache-if-present, then remote, then ongoing updates. `WaitForRemote(timeout)` = block the initial on the remote up to `timeout`, else cache fallback, else `(null, REMOTE, null)` — then ongoing updates, the same as `All`. **Any** successful fetch (subscribe- or `updateConfigs()`-triggered) delivers the latest config to **all** current subscribers, de-duplicated per subscriber by `fetchId`; failures are never broadcast (subscribers keep their last valid config); `updateConfigs()` is rate-limited (5 min). Never a stale `CACHE` after a `REMOTE`; storage persistence is best-effort. Mirrors iOS (incl. `AmplitudeCore-Swift#80`).
- **Cache:** an empty cached blob is a valid hit; stale old flat-format blobs (top-level keys containing `.`) are ignored and overwritten by a successful fetch, not migrated.
- **Gating:** `RemoteConfigClient`, its nested types, the `ConfigMap` extensions, and `Amplitude.remoteConfigClient` are behind `@RestrictedAmplitudeFeature` (ERROR-level opt-in) — an internal mechanism, consumed via narrowly-scoped `@OptIn`.

**Breaking (binary):** `Key` enum→sealed and `onUpdate` `long`→`Long?` are binary-incompatible, but confined to the now-gated internal `RemoteConfigClient`; see the package README migration note. No public-facing feature exposes these types.

- `./gradlew :core:test :android:test :core:apiCheck :android:apiCheck ktlintCheck` — all green.
- `amplitude-sdk-verification` remote-config scenarios (sealed key, delivery modes, `WaitForRemote` via MockWebServer incl. receiving a later refresh, dot-path, cache) pass against this branch.

- Package README: `core/src/main/java/com/amplitude/core/remoteconfig/README.md` (mermaid diagrams + migration note)
- iOS reference: `AmplitudeCore-Swift/Sources/AmplitudeCore/Remote Config/RemoteConfigClient.swift`; subsequent-update parity: `AmplitudeCore-Swift#80`

### Checklist
* [x] Does your PR have a breaking change? binary-incompatible, but confined to the gated internal \`RemoteConfigClient\` (see [README migration note](https://github.com/amplitude/Amplitude-Kotlin/blob/6ba6602282dbc293149516e6c3ef81504f9403a5/core/src/main/java/com/amplitude/core/remoteconfig/README.md#compatibility))

---

> [!NOTE]
> **High Risk**
> Large rewrite of remote-config fetch, cache, and callback delivery affects autocapture, network tracking, diagnostics, and session replay; binary-incompatible API changes require all internal subscribers to handle null config/timestamp correctly.
> 
> **Overview**
> **Aligns Android remote config with iOS/Web:** one nested blob from `config_group=android`, dot-path slices at delivery, and shared subscribe/delivery semantics.
> 
> **API and gating:** `RemoteConfigClient.Key` moves from an enum to a sealed class (`AnalyticsSdk`, `Diagnostics`, session-replay keys, `Custom("dot.path")`). `onUpdate` now takes nullable `config` and `timestamp` (null timestamp = fetch failed; null config with a timestamp = key missing). The client, `Amplitude.remoteConfigClient`, and `ConfigMap` helpers are behind `@RestrictedAmplitudeFeature`; internal callers opt in and use the new key names.
> 
> **Client behavior:** Subscribers share one in-flight fetch with `fetchId` de-duplication; successful refreshes broadcast to all subscribers while failures are not fan-out. Delivery modes only control the **initial** callback—after that, everyone gets updates. Cache vs remote ordering is fixed (no stale cache after remote), old flat cached blobs are ignored, persistence is best-effort, and `WaitForRemote` matches iOS (including later refreshes after the first delivery).
> 
> **Android consumers:** `AutocaptureManager` and `NetworkTrackingPlugin` handle nullable config (network tracking resets to local options on null). Tests and `core.api` reflect the binary API change; a package README documents behavior and migration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8868a65c94cc357ce43121c49560626c62dc5e31. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -962,7 +962,6 @@ public final class com/amplitude/core/remoteconfig/RemoteConfigClient$DeliveryMo
 }
 
 public abstract class com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
-	public static final field Companion Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key$Companion;
 	public synthetic fun <init> (Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public fun getValue ()Ljava/lang/String;
 }
@@ -972,9 +971,6 @@ public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$Analyt
 	public fun equals (Ljava/lang/Object;)Z
 	public fun hashCode ()I
 	public fun toString ()Ljava/lang/String;
-}
-
-public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$Companion {
 }
 
 public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$Custom : com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -961,15 +961,49 @@ public final class com/amplitude/core/remoteconfig/RemoteConfigClient$DeliveryMo
 	public final fun getTimeoutMs ()J
 }
 
-public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key : java/lang/Enum {
-	public static final field ANALYTICS_SDK Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key;
-	public static final field DIAGNOSTICS Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key;
-	public static final field SESSION_REPLAY_PRIVACY_CONFIG Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key;
-	public static final field SESSION_REPLAY_SAMPLING_CONFIG Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key;
-	public static fun getEntries ()Lkotlin/enums/EnumEntries;
-	public final fun getValue ()Ljava/lang/String;
-	public static fun valueOf (Ljava/lang/String;)Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key;
-	public static fun values ()[Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key;
+public abstract class com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
+	public static final field Companion Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key$Companion;
+	public synthetic fun <init> (Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun getValue ()Ljava/lang/String;
+}
+
+public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$AnalyticsSdk : com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
+	public static final field INSTANCE Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key$AnalyticsSdk;
+	public fun equals (Ljava/lang/Object;)Z
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$Companion {
+}
+
+public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$Custom : com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
+	public fun <init> (Ljava/lang/String;)V
+	public fun equals (Ljava/lang/Object;)Z
+	public fun getValue ()Ljava/lang/String;
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$Diagnostics : com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
+	public static final field INSTANCE Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key$Diagnostics;
+	public fun equals (Ljava/lang/Object;)Z
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$SessionReplayPrivacyConfig : com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
+	public static final field INSTANCE Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key$SessionReplayPrivacyConfig;
+	public fun equals (Ljava/lang/Object;)Z
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$SessionReplaySamplingConfig : com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
+	public static final field INSTANCE Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Key$SessionReplaySamplingConfig;
+	public fun equals (Ljava/lang/Object;)Z
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
 }
 
 public abstract interface class com/amplitude/core/remoteconfig/RemoteConfigClient$RemoteConfigCallback {
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -958,7 +958,10 @@ public final class com/amplitude/core/remoteconfig/RemoteConfigClient$DeliveryMo
 
 public final class com/amplitude/core/remoteconfig/RemoteConfigClient$DeliveryMode$WaitForRemote : com/amplitude/core/remoteconfig/RemoteConfigClient$DeliveryMode {
 	public fun <init> (J)V
+	public fun equals (Ljava/lang/Object;)Z
 	public final fun getTimeoutMs ()J
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
 }
 
 public abstract class com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -1002,7 +1002,7 @@ public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$Sessio
 }
 
 public abstract interface class com/amplitude/core/remoteconfig/RemoteConfigClient$RemoteConfigCallback {
-	public abstract fun onUpdate (Ljava/util/Map;Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Source;J)V
+	public abstract fun onUpdate (Ljava/util/Map;Lcom/amplitude/core/remoteconfig/RemoteConfigClient$Source;Ljava/lang/Long;)V
 }
 
 public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Source : java/lang/Enum {
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -963,7 +963,7 @@ public final class com/amplitude/core/remoteconfig/RemoteConfigClient$DeliveryMo
 
 public abstract class com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
 	public synthetic fun <init> (Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public fun getValue ()Ljava/lang/String;
+	public final fun getValue ()Ljava/lang/String;
 }
 
 public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$AnalyticsSdk : com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
@@ -976,7 +976,6 @@ public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$Analyt
 public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key$Custom : com/amplitude/core/remoteconfig/RemoteConfigClient$Key {
 	public fun <init> (Ljava/lang/String;)V
 	public fun equals (Ljava/lang/Object;)Z
-	public fun getValue ()Ljava/lang/String;
 	public fun hashCode ()I
 	public fun toString ()Ljava/lang/String;
 }
--- a/core/src/main/java/com/amplitude/core/remoteconfig/README.md
+++ b/core/src/main/java/com/amplitude/core/remoteconfig/README.md
@@ -30,16 +30,19 @@ iOS (`AmplitudeCore-Swift`) and Web clients so behavior is consistent across SDK
 
 ## Delivery semantics
 
-- **`subscribe` always delivers exactly once** initially — cache, remote, or a
-  failure signal — so callers can reliably gate setup on it.
+- **`subscribe` always delivers an initial callback** — cache, remote, or a failure
+  signal — so callers can reliably gate setup on it. The mode controls only this
+  *initial* delivery; afterward every subscriber receives subsequent successful
+  fetches as updates.
 - **`All`**: delivers cache immediately (if present), then the remote result, then
-  every subsequent successful refresh.
-- **`WaitForRemote(timeout)`**: waits for the remote up to `timeout`; on
-  failure/timeout falls back to cache, else a failure signal. **One-and-done** —
-  later refreshes do not re-notify it.
-- **`updateConfigs()`** (refresh): rate-limited (5 min). Only *successful* configs
-  are delivered to existing subscribers; failures are never fanned out, so
-  subscribers keep their last valid config.
+  every subsequent successful fetch.
+- **`WaitForRemote(timeout)`**: blocks the *initial* delivery on the remote up to
+  `timeout`; on failure/timeout falls back to cache, else a failure signal. After
+  that initial callback it receives subsequent fetches as updates, the same as `All`.
+- **Any successful fetch** — whether triggered by a `subscribe` or by
+  `updateConfigs()` — delivers the latest config to **all** current subscribers
+  (de-duplicated per subscriber by `fetchId`). Failures are never fanned out, so
+  subscribers keep their last valid config. `updateConfigs()` is rate-limited (5 min).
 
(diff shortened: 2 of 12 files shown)


polbins:
bugbot run

polbins:
bugbot run

polbins:
bugbot run

polbins:
@cursor push 3f53c2361d

polbins:
Marked as draft again, found an issue with the API dump. Will re-review and mark as ready again when done addressing that.

polbins:
bugbot run

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -408,95 +399,104 @@ internal class RemoteConfigClientImpl(
...
-                        }
-                    }
-                    return null
-                }
-                // If at least one expected key exists, keep partial data and proceed.
-            }
+            val blob = getStoredConfigBlob() ?: return null
 
-            val configData = allStoredConfigs[configKey] ?: emptyMap()
+            val resolved = resolveDotPath(blob, dotPath) ?: emptyMap()
```
sojingle: Minor parity nit, not blocking — when a subscribed dot-path is absent from the fetched blob, this implementation delivers emptyMap() to the callback, while both Swift and TS deliver null instead. The Kotlin choice is arguably more ergonomic (callers never have to null-check before using getBoolean / getString helpers), but subscribers can no longer distinguish "the key is missing from the server response" from "the key exists but is {}".

sojingle (approved):
Overall LGTM

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
polbins: Ahh, good catch on parity. Keeping `emptyMap()` for now since all current subscribers treat absent and empty the same (also saves us from null checks), but will revisit if a Custom key use case needs the distinction.

core/api/core.api
```diff
@@ -961,15 +961,49 @@ public final class com/amplitude/core/remoteconfig/RemoteConfigClient$DeliveryMo
 	public final fun getTimeoutMs ()J
 }
 
-public final class com/amplitude/core/remoteconfig/RemoteConfigClient$Key : java/lang/Enum {
```
crleona, line 964: any current consumers of this?

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
crleona: do we have a way for clients to make this distinction?

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -187,21 +211,23 @@ internal class RemoteConfigClientImpl(
...
 
-    // Simple in-flight fetch guard; safe because networkIODispatcher is single-threaded
+    // Additional root keys registered by Custom key subscribers, unioned with fetchKeys for the URL.
+    private val customFetchRoots = mutableSetOf<String>()
+
+    @Volatile
     private var isFetching: Boolean = false
 
+    // Set when customFetchRoots grows — bypasses rate limit on next fetch so
+    // newly registered custom roots get fetched promptly, not after 5 minutes.
```
crleona: We don't want this behavior. Keys should be as consistent as possible to increase cache hits at the CDN layer. We never want to refetch for new keys - we would just update core and increase the minimum version of the new dependency with this setup.

A better solution may be to just adopt config groups now, browser should use them, it looks like I didn't merge them yet for Swift. https://github.com/amplitude/AmplitudeCore-Swift/pull/17/changes

crleona: (this lets us update the fetched keys by dynconf)

core/api/core.api
polbins: No, it's just diagnostics and it's within this repo.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
polbins: Not right now, custom keys are `@RestrictedAmplitudeFeature` and internal-only, so absent vs empty hasn't been a real need. Keeping `emptyMap()` avoids null check for every subscriber, if a future internal use case needs the distinction we can switch to `ConfigMap?` without breaking external consumers.

crleona: I'd like to have this reflect what we believe the final APIs will be (as this will be used by SR, GS, etc and we will need to maintain binary compatibility) - it seems like we would want to be able to distinguish this case for consumers.

polbins: Good point. I was trying to keep the subscriber API simple by always passing a non-null map. Since current consumers only use default config when remote config is missing or empty, I did not see a need for nullable yet. But if we expect SR or GS to depend on this distinction, I’m happy to adjust it now.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -408,95 +368,104 @@ internal class RemoteConfigClientImpl(
...
-                            logger.debug(
-                                "Skipping non-map value for key $key: ${value?.javaClass?.simpleName}",
-                            )
-                        }
+            val allConfigs = JSONObject(configJson).toMapObj().filterNotNullValues()
+            if (allConfigs.isEmpty()) return null
+
+            // Detect old pre-flattened format: keys contain dots (e.g. "sessionReplay.sr_android_privacy_config").
+            // New format has top-level keys without dots (e.g. "sessionReplay", "diagnostics").
+            val isOldFormat = allConfigs.keys.any { "." in it }
```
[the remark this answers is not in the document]
polbins: Temporary migration code

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -408,95 +369,104 @@ internal class RemoteConfigClientImpl(
...
```
crleona, line 577: wondering if we should just drop any old format configs and wait for a new copy from remote?

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -52,11 +52,27 @@ interface RemoteConfigClient {
...
-        ANALYTICS_SDK("analyticsSDK.androidSDK"),
-        DIAGNOSTICS("diagnostics.androidSDK"),
-        SESSION_REPLAY_PRIVACY_CONFIG("sessionReplay.sr_android_privacy_config"),
-        SESSION_REPLAY_SAMPLING_CONFIG("sessionReplay.sr_android_sampling_config"),
+    /**
+     * Dot-path key for subscribing to remote config sections.
+     * Split by "." to walk nested config objects. For example,
+     * "sessionReplay.sr_android_privacy_config" resolves configs.sessionReplay.sr_android_privacy_config.
+     */
+    sealed class Key(open val value: String) {
```
crleona: is this still useful?

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
crleona: Do we give any response if the fetch fails? For clients that are sensitive to getting the latest remote config, they may wait for any response to complete setup. We should always send at least a null response to indicate this failed state. (iOS sends a null response from remote).

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
polbins: I think it makes sense to drop, applied those changes and it should match iOS behavior now

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
polbins: Yes, we're sending null as well

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -582,62 +545,34 @@ internal class RemoteConfigClientImpl(
...
-                    }
-
-                if (result.isEmpty() && configsRoot.length() == 0) {
-                    // The "configs" root is present but empty — project has no
-                    // blades enabled. This is a valid server answer; return an
-                    // empty map so that downstream logic (notifyAbsentKeySubscribers)
-                    // can unblock WaitForRemote subscribers immediately.
+                if (configsRoot.length() == 0) {
                     logger.debug("Server returned empty configs — project may have no blades enabled")
                     return emptyMap()
```
crleona, line 706: should this return null?

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -308,92 +309,73 @@ internal class RemoteConfigClientImpl(
...
 
-                isFetching = true
-                val configs = fetchRemoteConfig() ?: return@launch
-
-                // Snapshot old cache before overwriting — absent-key fallback
-                // needs the pre-fetch values, not the partial new response.
-                val preFetchCache = getAllStoredConfigs()
+                val blob = fetchRemoteConfig()
+                if (blob == null) {
+                    notifySubscribersOnFailure(System.currentTimeMillis())
```
crleona: At least on iOS, we only deliver failures if both cache and remote fail on subscribe, never from updates. 

For each call to subscribe, we want to ensure we always deliver a response, so clients can reliably wait for remote config before fully setting up / initializing. For this initial case, we can either return success from cache or remote, or an actual failure so the clients can handle the no config available case.

For subsequent updates, I don't think we'd ever want to return a failed response - we'd always want to use the latest valid config available, so we shouldn't have to notify subscribers - effectively nothing changes for them.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
polbins: The `open` keyword is not, we already provided a `Custom(value)` and making this open is not really needed. Fixing..

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -308,92 +309,73 @@ internal class RemoteConfigClientImpl(
...
-                    }
-                if (delivered) {
-                    logger.debug(
-                        "WaitForRemote unblocked for $configKey: key absent from fetch; " +
-                            "delivered ${if (cachedConfig != null) "cache" else "empty"} fallback.",
-                    )
+        for (subscribers in subscriberSnapshot) {
+            subscribers.forEach { weakCallback ->
+                weakCallback.runSafely {
+                    onUpdate(null, REMOTE, timestamp)
```
crleona: How would a client tell if a response is empty for their selected key path or if it's an error? on iOS, we don't return a timestamp for the error case (timestamp == time of successful fetch)

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -308,92 +309,73 @@ internal class RemoteConfigClientImpl(
...
 
+            // Claim the in-flight slot synchronously, before the suspending
+            // shouldRateLimit() read. Otherwise a second coroutine queued on the
+            // single network dispatcher could pass the isFetching guard during the
+            // suspension and issue a duplicate fetch.
+            isFetching = true
+            try {
                 if (shouldRateLimit()) {
                     logger.debug("RemoteConfig update skipped: within 5-minute window")
                     return@launch
```
crleona: If we subscribe with no cached data, would we be guaranteed to get a callback if we're rate limited?

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
polbins: Ahh, that makes sense. I'll make the timestamp nullable, so there will be error (null) and absent key (valid timestamp).

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
polbins: We usually return `null` for invalid, but this one seems valid but just empty.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
polbins: This is silent for `All` without cache, and `WaitForRemote` will wait for timeout. I'll fill up this gap 👍🏽

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -214,72 +272,101 @@ internal class RemoteConfigClientImpl(
...
+            when (val handle = obtainSubscribeFetch()) {
+                is FetchHandle.Reuse ->
+                    weakCallback.deliverRemote(
+                        resolveDotPath(handle.success.blob, key.value),
+                        handle.success.timestamp,
+                        handle.success.fetchId,
+                    )
+                is FetchHandle.Pending ->
+                    when (val outcome = handle.deferred.await()) {
+                        is FetchOutcome.Success ->
```
crleona, line 313: should a successful fetch broadcast updated configs to all subscribers?

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
```diff
@@ -98,109 +140,125 @@ internal class RemoteConfigClientImpl(
...
-            if (!claimed) return false
-
-            firstDeliverySignal?.complete(Unit)
             synchronized(deliveryLock) {
-                try {
-                    callback.action()
-                } catch (e: Exception) {
-                    logger.error("Exception in subscriber callback: ${e.message}")
-                }
+                if (!delivered.compareAndSet(false, true)) return false
```
crleona, line 229: it looks like all waitforremote subscribers go through this call - does this mean that subsequent updates won't be delivered? If I call `updateConfigs` and get a success response after the initial waitforremote response is delivered, shouldn't I get a callback?

core/src/main/java/com/amplitude/core/remoteconfig/README.md
```diff
@@ -0,0 +1,111 @@
+# Remote Config
+
+Fetches a nested configuration blob from Amplitude's remote-config service and
```
crleona, line 3: nice, this is a good spec to work off of

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
polbins: This seems to align with iOS: `WaitForRemote` is one-and-done, `updateConfigs()` only re-delivers before it's first callback fires. 

The KDoc in this PR needs to be updated though, it currently still states :
- `Subsequent updates after the first delivery always flow through normally.` 

but it should be updated to 
- `a successful [updateConfigs] fans out to [All] subscribers always, but to a [WaitForRemote] subscriber only while it is still awaiting its first callback.`

Let me know which way we prefer, the DeliveryMode started with a simple "how the first config is delivered", but it should also be how **initial** config is delivered and whether it receives later **refreshes**.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
polbins: If it catches a fetch in flight, it self-delivers to that subscriber only. Broadcast to all current subscribers only happens via `updateConfigs()`.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
crleona: This is a bug in iOS. We'd always want to deliver the latest values if available.

core/src/main/java/com/amplitude/core/remoteconfig/RemoteConfigClient.kt
crleona: why wouldn't we update on any successful fetch? We should use the latest available configs.

polbins: The case I had in mind for not refreshing mid-session is something for `Experiment` where we typically resolve config once at session start and keep it stable for the session for consistency, but of course always use the latest can be the intended semantic. Changing it to use that 👍🏽

crleona (approved):

PR #410, opened 2026-06-02
feat(core): Plugin contract — state callbacks, identity notifications, dedup (SDKA-6)

polbins:

Blade SDKs (Session Replay, G&S, Experiment) and third-party hosts (Segment, mParticle) need to observe identity and lifecycle changes on the Amplitude instance they're plugged into. Before this PR, `Plugin` had no callbacks for userId/deviceId/optOut/reset/session changes, no way to find a registered plugin by type, and no deduplication when re-adding the same plugin. Reset was also broken — it fired two separate identity notifications (userId then deviceId) instead of one atomic wave.

### Describe the solution

**`IdentityCoordinator` returns, doesn't dispatch.** `setUserId`, `setDeviceId`, `resetIdentity`, and `bootstrap` each return an `EnumSet<State.IdentityChange>` describing what changed and notify nothing. `Amplitude` is the single dispatch owner — it turns that return value into plugin callbacks. This keeps notification structurally outside the identity lock (reentrancy fix), with no dual-path surface that could cause double-fires or ordering bugs.

**`Plugin` interface additions** (all defaulted — binary compat preserved):
- `onUserIdChanged(userId)`, `onDeviceIdChanged(deviceId)`, `onOptOutChanged(optOut)`, `onReset()`, `onSessionIdChanged(sessionId)`
- `name: String?` — optional identifier for replace-on-add dedup (teardown called on the displaced plugin)

**`Amplitude` additions:**
- `findPlugin<T>()` — searches timeline and observe store
- `notifyAllPlugins(block)` — single snapshot fan-out with per-plugin try/catch (one bad plugin doesn't block others)
- `resetIdentity(newDeviceId)` — atomic identity reset, fires one `onReset` wave

**Android:**
- `AndroidContextPlugin.resolveDeviceId` split out from `initializeDeviceId` so `reset()` can resolve the new device id before calling `resetIdentity` — eliminates the double identity-change
- `Timeline` fires `onSessionIdChanged` on both session restore (persisted session) and new session start
- `Type.Observe` mediator wired into `Timeline` so bare `Plugin` with `type = Observe` is no longer silently dropped

```bash
./gradlew :core:test :android:test apiCheck
```

New test classes: `PluginContractTest` (core), `PluginContractAndroidTest` (Android) — cover fan-out, dedup, reset atomicity, session callback on restore, and named-plugin replace-on-add.

[SDKA-6 — Linear](https://linear.app/amplitude/issue/SDKA-6/plugin-contract-state-callbacks-dedup-findplugin)

### Checklist

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -133,12 +133,14 @@ open class Amplitude internal constructor(
     }
 
     override fun reset(): Amplitude {
-        setUserId(null)
-        if (isBuilt.isCompleted) {
-            androidContextPlugin.initializeDeviceId(configuration as Configuration, forceRegenerate = true)
-        } else {
-            setDeviceId(ContextPlugin.generateRandomDeviceId())
-        }
+        val newDeviceId =
+            if (isBuilt.isCompleted) {
+                androidContextPlugin.resolveDeviceId(configuration as Configuration, forceRegenerate = true)
+                    ?: ContextPlugin.generateRandomDeviceId()
+            } else {
+                ContextPlugin.generateRandomDeviceId()
+            }
+        resetIdentity(newDeviceId)
         return this
     }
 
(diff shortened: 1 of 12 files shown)


polbins:
bugbot run

polbins:
@cursor push 67eeaef651

PR #412, opened 2026-06-04
feat: plugin contract for blades - state callbacks, identity notifications

polbins:

Part of the Unified SDK work (**SDKA-6**). "Blades" — Session Replay, Experiment, Guides & Surveys — need to register as plugins on the analytics `Amplitude` instance and react to identity/session/opt-out changes. Today the plugin model only delivers `onUserIdChanged`/`onDeviceIdChanged`, and only to `ObservePlugin`s in the `State` store — timeline plugins and the other state transitions (session, opt-out, reset) aren't surfaced, and there's no way to look a plugin up or dedup one by name.

### Describe the solution

A small, additive plugin contract layered on top of the existing pipeline — no rewrite of identity internals.

- **`Plugin` gains five default no-op callbacks** — `onUserIdChanged`, `onDeviceIdChanged`, `onSessionIdChanged` (Android), `onOptOutChanged`, `onReset` — so existing plugins are unaffected.
- **`findPlugin<T>()`** across the timeline mediators **and** the observe store, and **name-based dedup** on `add()` (a named plugin replaces + tears down the previous).
- **One fan-out path** (`Amplitude.notifyPlugins`): snapshots timeline + store once and isolates each call (`safelyNotify`), so a throwing plugin can't break the loop, the calling thread, or the Android event-message coroutine. `reset()` reuses a single snapshot for its userId → deviceId → reset passes.

Design decisions / trade-offs:
- **`com.amplitude.id.*` is untouched.** It's shared with the Experiment SDK and is the binary-compat boundary — identity persistence stays where it is. The shipped `IdentityCoordinator` remains the single runtime identity owner (lock + `State` mirror + pre-build pending + bootstrap reconcile); it gains an atomic `reset(newDeviceId)`.
- **Reentrancy-safe by construction.** `State.userId/deviceId` became silent `@Volatile` mirrors; all notification moved to `Amplitude` and fires **after** the identity lock is released — so a plugin that re-sets identity from inside its own callback isn't clobbered by the outer commit. No lock is ever held across a plugin callback.
- Notification is **intent-based** (every explicit set notifies once), one rule for all five callbacks.

Incidentally fixes two latent issues on `main`: a cross-thread visibility gap on the (previously non-`@Volatile`) identity fields, and userId/deviceId tearing in Android `reset()` (previously two separate mutations, now one atomic reset).

**Public API:** additive only — `core.api` +8 (`Plugin` name + 5 callbacks, `protected doResetWithDeviceId`/`notifyAllPlugins`); `android.api` unchanged. **Not a breaking change.**

- `./gradlew :core:test :android:test` — green. Core includes a new `PluginContractTest` (179 cases incl. reentrancy + `@RepeatedTest` concurrency) and Android `PluginContractAndroidTest` (session/reset/dedup).
- `./gradlew apiCheck` — green; `core/api/core.api` is +8 additive, `android/api/android.api` unchanged.
- **sdk-verification harness** — verified against this branch under a customer-shaped Android/Gradle/AGP/Kotlin setup; all plugin/remote-config/remote scenarios pass, including a new `PluginConcurrencyScenario` (reentrant `setUserId`/`setDeviceId` from a callback; concurrent add/setUserId/reset with no deadlock). The scenario coverage lands as a **follow-up PR in the `amplitude-sdk-verification` repo** (currently on a branch); run locally with `cd android-baseline && ./gradlew :scenarios:testDebugUnitTest -Pamplitude.sdk.path=<this-clone>`.

- SDKA-6 (Jira) · Unified SDK for Android (Confluence)
- Plugin & identity contributor guide added in `core/src/main/java/com/amplitude/core/README.md`
- Follow-up: `amplitude-sdk-verification` PR adding `PluginConcurrencyScenario` + realigning the reset scenario to the shipped API.

### Checklist
* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-Kotlin/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)? (`feat:`)
* [ ] Does your PR have a breaking change? — **No** (additive public API only)

---

> [!NOTE]
> **Medium Risk**
> Touches core identity mutation and plugin notification ordering (reset, reentrancy, session restore); behavior is additive but blades will depend on these semantics.
> 
> **Overview**
> Adds a **unified plugin contract** so blades can react to identity and SDK state without living only in the deprecated `State` observe registry.
> 
> **`Plugin`** now exposes optional **`name`** plus default no-op callbacks (`onUserIdChanged`, `onDeviceIdChanged`, `onSessionIdChanged`, `onOptOutChanged`, `onReset`). **`Amplitude`** fans these out through one **`notifyPlugins`** path: snapshot registered plugins once, **`safelyNotify`** per plugin, and fire **after** `IdentityCoordinator` releases its lock so reentrant `setUserId` / `setDeviceId` from a callback is not clobbered. **`reset()`** goes through **`doResetWithDeviceId`** and **`IdentityCoordinator.reset`**, delivering userId → deviceId → `onReset` from a single snapshot. **`optOut`** assignment on `Amplitude` also notifies plugins.
> 
> **`State`** identity fields are silent **`@Volatile`** mirrors (no notification in setters); the legacy **`State.plugins`** registry is deprecated. **`Timeline`** gains an **Observe** mediator, **`findPlugin<T>()`**, and **name-based dedup** (first registration wins). **Android** calls **`notifySessionIdChanged`** when sessions are restored, configured, or rotated; **`reset()`** resolves a new device id via **`AndroidContextPlugin.resolveDeviceId`** then uses the shared reset path.
> 
> Contributor docs in **`core/README.md`** and broad **`PluginContractTest`** / **`PluginContractAndroidTest`** coverage accompany additive **`core.api`** surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5deb61577782f7490ec2e2a97102dd0d7cebd0ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/core/src/main/java/com/amplitude/core/Amplitude.kt
+++ b/core/src/main/java/com/amplitude/core/Amplitude.kt
@@ -43,7 +43,6 @@ import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.launch
 import java.io.File
-import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.Executors
 
 /**
@@ -60,7 +59,6 @@ open class Amplitude(
     val storageIODispatcher: CoroutineDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher(),
 ) {
     val timeline: Timeline
-    private val reservedPlugins: ConcurrentHashMap<String, Plugin> = ConcurrentHashMap()
     val storage: Storage by lazy {
         configuration.storageProvider.getStorage(this)
     }
@@ -143,6 +141,7 @@ open class Amplitude(
     init {
         require(configuration.isValid()) { "invalid configuration" }
         timeline = this.createTimeline()
+        store.owner = this
         isBuilt = this.build()
         isBuilt.start()
     }
@@ -544,42 +543,17 @@ open class Amplitude(
      * @return the Amplitude instance
      */
     fun add(plugin: Plugin): Amplitude {
-        val name = plugin.name
-        if (name != null && reservedPlugins.putIfAbsent(name, plugin) != null) {
-            logger.warn("Plugin \"$name\" is already registered; keeping the existing one.")
-            return this
-        }
-        try {
-            @Suppress("DEPRECATION")
-            when (plugin) {
-                is ObservePlugin -> store.add(plugin, this)
-                else -> timeline.add(plugin)
-            }
-        } catch (t: Throwable) {
-            if (name != null) reservedPlugins.remove(name, plugin)
-            throw t
-        }
+        timeline.add(plugin)
         return this
     }
 
     /**
-     * Find the first registered plugin assignable to [T], across both the timeline mediators
-     * and the observe store (store iteration is snapshotted under its lock).
+     * Find the first registered plugin assignable to [T] in the timeline mediators.
      */
-    inline fun <reified T : Plugin> findPlugin(): T? {
-        timeline.findPlugin<T>()?.let { return it }
-        @Suppress("DEPRECATION")
-        val observeSnapshot = synchronized(store.plugins) { store.plugins.toList() }
-        return observeSnapshot.firstOrNull { it is T } as? T
-    }
+    inline fun <reified T : Plugin> findPlugin(): T? = timeline.findPlugin<T>()
 
     fun remove(plugin: Plugin): Amplitude {
-        plugin.name?.let { reservedPlugins.remove(it, plugin) }
-        @Suppress("DEPRECATION")
-        when (plugin) {
-            is ObservePlugin -> store.remove(plugin)
-            else -> timeline.remove(plugin)
-        }
+        timeline.remove(plugin)
         return this
     }
 
@@ -603,7 +577,9 @@ open class Amplitude(
         pluginsSnapshot().forEach { safelyNotify(it, block) }
     }
 
-    private fun pluginsSnapshot(): List<Plugin> = timeline.pluginsSnapshot() + store.pluginsSnapshot()
+    private fun pluginsSnapshot(): List<Plugin> = timeline.pluginsSnapshot()
+
+    internal fun observePluginsSnapshot(): List<ObservePlugin> = timeline.pluginsSnapshot().filterIsInstance<ObservePlugin>()
 
     private fun safelyNotify(
         plugin: Plugin,
--- a/core/src/main/java/com/amplitude/core/Amplitude.kt
+++ b/core/src/main/java/com/amplitude/core/Amplitude.kt
@@ -133,6 +133,7 @@ open class Amplitude(
         get() = configuration.optOut
         set(value) {
             configuration.optOut = value
+            notifyPlugins { it.onOptOutChanged(value) }
         }
 
     init {
@@ -307,6 +308,7 @@ open class Amplitude(
      */
     fun setUserId(userId: String?): Amplitude {
         identityCoordinator.setUserId(userId)
+        notifyPlugins { it.onUserIdChanged(store.userId) }
         return this
     }
 
@@ -327,6 +329,7 @@ open class Amplitude(
      */
     fun setDeviceId(deviceId: String): Amplitude {
         identityCoordinator.setDeviceId(deviceId)
+        notifyPlugins { it.onDeviceIdChanged(store.deviceId) }
         return this
     }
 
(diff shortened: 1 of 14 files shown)


polbins:
bugbot run

core/src/main/java/com/amplitude/core/Amplitude.kt
sojingle: `onOptOutChanged` only notifies in this setter, won't be **equivalent** with Configuration.optOut

sojingle (approved):

core/src/main/java/com/amplitude/core/Amplitude.kt
polbins: Yes, this is by design. We've been moving away from direct `configuration` access, and `Amplitude.optOut` is the preferred path going forward especially since it's what notifies plugins.

I'll update the KDoc to stop claiming equivalence ("mutating either is equivalent") and make that more explicit. Let me know if that's suffice.

sojingle: Sounds good. Thanks!

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -517,6 +542,10 @@ open class Amplitude(
      * @return the Amplitude instance
      */
     fun add(plugin: Plugin): Amplitude {
+        // Dedup by name (best-effort; add() is expected from a single thread): a named plugin
```
crleona: we should make this thread safe

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -517,6 +542,10 @@ open class Amplitude(
...
+        // replaces and tears down any existing plugin with the same name.
+        plugin.name?.let { removePluginsByName(it) }
```
crleona: iOS we don't replace, the first plugin persists. Could you double check the behavior with web as well?

core/src/main/java/com/amplitude/core/platform/Timeline.kt
```diff
@@ -10,6 +10,7 @@ open class Timeline {
             Plugin.Type.Enrichment to Mediator(),
             Plugin.Type.Destination to Mediator(),
             Plugin.Type.Utility to Mediator(),
+            Plugin.Type.Observe to Mediator(),
```
crleona, line 14: it doesnt seem like this is used. We should be able to treat observe plugins like any other now that every plugin receives device/user id updates, and remove the plugin logic from State.

core/src/main/java/com/amplitude/core/Amplitude.kt
polbins: Good call, checked web too and it also keeps the first plugin. Updated this to match that behavior and added a warning.

core/src/main/java/com/amplitude/core/platform/Timeline.kt
polbins: This is for bare plugins that use `type = Plugin.Type.Observe` and do not extend `ObservePlugin`, so this is mostly for behavior parity just in case someone is using it. (there's a test named `bare observe type plugin is stored in the timeline`)

As for `State`, agreed on the direction but `State.add/remove/plugins` are pre-existing public API. 

I've deprecated them and pointed callers to Amplitude.add/remove, lmk if you prefer a different approach.

core/src/main/java/com/amplitude/core/State.kt
```diff
@@ -2,35 +2,69 @@ package com.amplitude.core
...
+                plugins.removeAll { it === plugin }
+            }
+        // teardown() runs outside the monitor — never hold a lock across plugin code.
+        if (removed) plugin.teardown()
+        return removed
     }
 
-    fun remove(plugin: ObservePlugin) =
+    @Suppress("DEPRECATION")
+    internal fun pluginsSnapshot(): List<ObservePlugin> =
```
crleona: do we need to add `pluginsSnapshot` if we're deprecating the plugin functionality here?

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -517,28 +544,42 @@ open class Amplitude(
...
-                this.timeline.add(plugin)
+        val name = plugin.name
+        if (name != null && reservedPlugins.putIfAbsent(name, plugin) != null) {
+            logger.warn("Plugin \"$name\" is already registered; keeping the existing one.")
+            return this
+        }
+        try {
+            @Suppress("DEPRECATION")
+            when (plugin) {
+                is ObservePlugin -> store.add(plugin, this)
```
crleona: Why do we still need to store the observeplugin separately in store? Is it functionally different to have it hosted under Timeline like any other plugin? I don't think compatibility is a strong enough argument for the additional complexity.

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -59,6 +60,7 @@ open class Amplitude(
     val storageIODispatcher: CoroutineDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher(),
 ) {
     val timeline: Timeline
+    private val reservedPlugins: ConcurrentHashMap<String, Plugin> = ConcurrentHashMap()
```
crleona: does this need to store a plugin or just the name to dedup?

core/src/main/java/com/amplitude/core/Amplitude.kt
polbins: > Why do we still need to store the observeplugin separately in store? Is it functionally different to have it hosted under Timeline like any other plugin?

This is the part where I'm not entirely sure, it was implemented as-is and `ObservePlugin` abstract class is handled differently. We did improve our identity handling recently, so I'll give this a go and see if we can eliminate that special case for Observe so we're all handling it on Timeline.

core/src/main/java/com/amplitude/core/Amplitude.kt
polbins: Moved this dedup into the Timeline (reservedNames) with the consolidated approach. It maps name→plugin, so remove frees the slot by identity and only the instance actually registered under that name releases it.

core/src/main/java/com/amplitude/core/Amplitude.kt
polbins: Updated, plugin storage is now consolidated into the Timeline.

Added `State.owner` which is a back-reference so we can keep the previous public API behavior running for the meantime. We can remove this in the future so that we don't have to maintain it.

core/src/main/java/com/amplitude/core/State.kt
```diff
@@ -2,35 +2,66 @@ package com.amplitude.core
...
     fun add(
         plugin: ObservePlugin,
         amplitude: Amplitude,
-    ) = synchronized(plugins) {
-        plugin.setup(amplitude)
-        plugins.add(plugin)
+    ): Boolean {
+        amplitude.add(plugin)
+        // Reflect the real outcome: first-wins name dedup may have skipped a duplicate.
+        return amplitude.observePluginsSnapshot().any { it === plugin }
```
crleona: I still would prefer to leave this code intact but deprecated vs trying to make a compatibility shim. Where observeplugins are stored is more an implementation detail than an intentional public API surface.

crleona (approved):

PR #419, opened 2026-07-14
feat(core): add host-agnostic plugin contract (SDKA-7)

polbins:

Blade SDKs need the same host-agnostic plugin contract available in `AmplitudeCore-Swift`, without depending on the concrete Android `Amplitude` client.

### Describe the solution

iOS-to-Android mapping:

| iOS | Android |
| --- | --- |
| `AnalyticsClient` | `AnalyticsClient` |
| `AnalyticsIdentity` | `AnalyticsIdentity` |
| `AnalyticsEvent` | `AnalyticsEvent` |
| `AmplitudeContext` | `AmplitudeContext` |
| `UniversalPlugin` | `UniversalPlugin` |
| `PluginHost.plugin(name:)` / `plugins(type:)` | `PluginHost.plugin(name)` / `plugins(type)` |

Existing Android `Plugin` implementations also conform to `UniversalPlugin`, and host-agnostic plugins are registered directly with `Amplitude` and hosted on the timeline without a wrapper.

Intentional divergences:

- **Client conformance:** Plugins receive a delegating `AnalyticsClient` instead of `Amplitude` conforming directly. This avoids a JVM return-type clash with the released fluent `track(...): Amplitude` API while preserving iOS-compatible `track(eventType, eventProperties)` behavior.
- **User properties:** The identity we hand to plugins exposes `userId`/`deviceId` only, so `userProperties` stays an empty map. User properties still work exactly as before — they're just not readable through this plugin-facing identity (iOS exposes them here).
- **Sessions:** The Android module supplies the live `sessionId`; host-agnostic core returns `-1` because session tracking lives in the Android layer.
- **Event execution:** Cross-host plugins operate on a minimal `AnalyticsEvent` (same shape as iOS), not the concrete `BaseEvent`. The Kotlin hook returns the event or `null` to drop rather than iOS's in-place `inout` — same modify/drop behavior, different shape. The internal Android pipeline still runs on `BaseEvent`.

All other lifecycle/lookup semantics mirror iOS.

```bash
./gradlew :core:test
./gradlew :android:test
./gradlew apiCheck
./gradlew ktlintCheck
./gradlew build -x test
```

- [SDKA-7](https://linear.app/amplitude/issue/SDKA-7/universalplugin-interface)
- [iOS `UniversalPlugin`](https://github.com/amplitude/AmplitudeCore-Swift/blob/main/Sources/AmplitudeCore/Plugins/UniversalPlugin.swift) and [`AnalyticsClient`](https://github.com/amplitude/AmplitudeCore-Swift/blob/main/Sources/AmplitudeCore/Plugins/AnalyticsClient.swift)
- [iOS `Amplitude` host conformance](https://github.com/amplitude/Amplitude-Swift/blob/main/Sources/Amplitude/Amplitude.swift)

### Checklist


polbins:
bugbot run

polbins:
bugbot run

polbins:
bugbot run

core/src/main/java/com/amplitude/core/platform/UniversalPlugin.kt
```diff
@@ -0,0 +1,58 @@
...
+    fun setup(
+        client: AnalyticsClient,
+        context: AmplitudeContext,
+    ) {}
+
+    /**
+     * Processes an event as it flows through the pipeline. Return the event (optionally
+     * modified) to let it continue, or `null` to drop it.
+     */
+    fun execute(event: BaseEvent): BaseEvent? = event
```
crleona: Universal plugins should be able to work with non-amplitude Events and SDKs. Would prefer to use something a bit simpler than BaseEvent that we can bridge vs having a bunch of fields that won't have data. Some other Analytics SDKs also don't have a way to block or replace the event, only to modify it (see mParticle: https://github.com/mParticle/mparticle-android-sdk/blob/6c2079ea15596744209000a1d1a7ddfcdced86ee/android-kit-base/src/main/java/com/mparticle/kits/KitIntegration.java#L487)

core/src/main/java/com/amplitude/core/platform/UniversalPluginAdapter.kt
```diff
@@ -0,0 +1,33 @@
+package com.amplitude.core.platform
+
+import com.amplitude.core.Amplitude
+import com.amplitude.core.AmplitudeContext
+import com.amplitude.core.AnalyticsClient
+import com.amplitude.core.AnalyticsIdentity
+import com.amplitude.core.events.BaseEvent
+
+/** Adapts a [UniversalPlugin] so it can be hosted by [Amplitude] on the standard [Plugin] timeline. */
+internal class UniversalPluginAdapter(val delegate: UniversalPlugin) : Plugin {
```
crleona: if plugin is a universalPlugin, why do we need an adaptor?

core/src/main/java/com/amplitude/core/Amplitude.kt
```diff
@@ -604,6 +685,26 @@ open class Amplitude(
     }
 }
 
+private class AmplitudeAnalyticsClient(
```
crleona, line 675: do you need a wrapper object?

core/src/main/java/com/amplitude/core/platform/UniversalPluginAdapter.kt
polbins: Since `Plugin : UniversalPlugin`, those just skip the need for the adapter.

But for bare `UniversalPlugin` it's needed as timeline runs on Plugin, which needs a type (for Before/Enrichment/Destination ordering) and an `amplitude` handle. The adapter also supplies them and pins the plugin to Enrichment. 

Instead of this approach, I could also teach `timeline` to accept UniversalPlugin directly. Happy to do that, if it's preferred to an adapter/wrapper.

core/src/main/java/com/amplitude/core/Amplitude.kt
polbins: Yes, though I wish we didn't. 

Unlike iOS, Amplitude can't conform to AnalyticsClient directly here. 
Our released `track(...): Amplitude` clashes with the interface's `track(...): Unit` on the JVM (same descriptor, different return type), and fixing it at the source would break binary compat. 

This small delegating client circumvents that. (I played with the idea of a different name briefly, like `trackEvent`, but this seems like a better approach so we can keep the original function name)

core/src/main/java/com/amplitude/core/platform/UniversalPluginAdapter.kt
crleona: Huh, the Amplitude reference on the Plugin interface is unique to Android. It doesn't look like it's used outside of `Plugin` - could we ignore it if we change the internal files to `UniversalPlugin`?  On iOS we converted everything to accept `UniversalPlugin`, we only needed to [special case the `type` logic](https://github.com/amplitude/Amplitude-Swift/blob/17ca168cfe6bc6400e0fd7d0287fb2c742ed0d21/Sources/Amplitude/Timeline.swift#L50) and the [initial setup call](https://github.com/amplitude/Amplitude-Swift/blob/17ca168cfe6bc6400e0fd7d0287fb2c742ed0d21/Sources/Amplitude/Amplitude.swift#L377). 

This seems like a simpler approach to me if we can also figure out handling for the `amplitude` var.

core/src/main/java/com/amplitude/core/Amplitude.kt
crleona: I don't hate the minor rename if it makes things easier (this is a new interface after all), but the wrapper works too for consistency. This is cleaner than the plugin wrapper.

core/src/main/java/com/amplitude/core/platform/UniversalPlugin.kt
polbins: Added `AnalyticsEvent` and execute now takes that instead of `BaseEvent`. `BaseEvent` implements it, so the internal pipeline is unchanged.

core/src/main/java/com/amplitude/core/platform/Timeline.kt
```diff
@@ -31,15 +31,30 @@ open class Timeline {
...
 
+    internal fun plugin(name: String): UniversalPlugin? = reservedNames[name]
+
     fun add(plugin: Plugin) {
+        register(plugin)
+    }
+
+    fun add(plugin: UniversalPlugin) {
+        register(plugin)
+    }
```
crleona: do you need both adds?

crleona (approved):

core/src/main/java/com/amplitude/core/platform/Timeline.kt
polbins: At first thought, I kept both adds because of binary compat. But thought about it more and it seems like this is actually an internal detail as clients don't usually add a plugin through timeline (only through `Amplitude`). Updated it to only have add/remove for `UniversalPlugin`.

PR #421, opened 2026-07-21
fix(android): stop autocapture after logout by fully tearing down the instance (SDK-171)

polbins:
Autocapture keeps firing after a user logs out and the app re-initializes with autocapture disabled. `flush()` + `reset()` + dropping the reference never tears the instance down: the constructor registers `ActivityLifecycleCallbacks` on the `Application` and never unregisters them, per-instance `Runtime` shutdown hooks are never removed, and the coroutine scope / executors keep running. So the old instance leaks and keeps autocapturing. iOS is unaffected. (SDK-171)

### Why an explicit teardown API (context for iOS / wider team)
iOS gets this for free through **ARC**: dropping the last reference deallocs the instance, and its `NotificationCenter` observers are zeroing-weak, so autocapture stops automatically — no teardown API needed. The JVM has **no ARC**, and the `Application` holds a **strong** reference to the registered lifecycle callbacks, so simply dropping the reference leaks the instance instead of tearing it down:

```
Android (today)   Application ──strong──▶ ActivityLifecycleCallbacks ──▶ Amplitude
                  drop reference ▶ still retained ▶ keeps autocapturing  ✗

iOS (ARC)         drop reference ▶ dealloc ▶ observers auto-removed ▶ stops  ✓
```

We evaluated automatic alternatives (weak self-unregistering callbacks; an instance registry that supersedes a same-name instance) — without ARC none are both complete and low-risk. The SDK can't reliably detect "this instance is done," so **the app has to signal it.** `shutdown()` is that signal. Practical takeaway: a single long-lived instance needs no change; **multi-instance / teardown-and-recreate usage must call `shutdown()`** before dropping the old instance.

### Describe the solution
Add a public `shutdown()` that permanently tears the instance down: unregister lifecycle + per-activity fragment callbacks, remove the `Runtime` shutdown hook, stop the timeline (tearing down every plugin), cancel `amplitudeScope`, and close the executor dispatchers. Idempotent (single-entry via `AtomicBoolean`); `track()`/`flush()` no-op afterwards; a `shutdown()` racing an in-flight async `build()` re-runs teardown. Stacked with #422, which frees the global EventBridge receiver so a same-`instanceName` rebuild works.

`./gradlew :analytics-core:test :android:test`. New tests assert no autocapture fires after `shutdown()`, `ActivityLifecycleCallbacks` are unregistered, and teardown is idempotent / safe when called before `build()` completes.

### Checklist
* [ ] Does your PR have a breaking change? No — additions-only public API (`apiCheck` passes).

### DIFF
--- a/analytics-core/api/analytics-core.api
+++ b/analytics-core/api/analytics-core.api
@@ -86,9 +86,11 @@ public class com/amplitude/core/Amplitude : com/amplitude/core/platform/PluginHo
 	public static synthetic fun identify$default (Lcom/amplitude/core/Amplitude;Lcom/amplitude/core/events/Identify;Lcom/amplitude/core/events/EventOptions;ILjava/lang/Object;)Lcom/amplitude/core/Amplitude;
 	public static synthetic fun identify$default (Lcom/amplitude/core/Amplitude;Ljava/util/Map;Lcom/amplitude/core/events/EventOptions;ILjava/lang/Object;)Lcom/amplitude/core/Amplitude;
 	public final fun isBuilt ()Lkotlinx/coroutines/Deferred;
+	protected final fun isShutdown ()Z
 	public final fun logEvent (Lcom/amplitude/core/events/BaseEvent;)Lcom/amplitude/core/Amplitude;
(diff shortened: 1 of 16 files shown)


polbins:
bugbot run

polbins:
bugbot run

PR #422, opened 2026-07-21
fix(core): clear global EventBridge receiver on shutdown so instances can be recreated (SDK-171)

polbins:
`EventBridgeContainer` keeps one global container per `instanceName`, and `EventBridge` only ever accepts a single receiver (once set it can't be replaced or cleared). So after an instance is shut down its slot leaks (retaining the instance), and rebuilding with the same `instanceName` silently drops the new instance's receiver — breaking cross-SDK / Experiment bridge events on the recreated instance. (SDK-171)

### Describe the solution
`shutdown()` now removes the container for its `instanceName`, freeing the slot so a same-named rebuild registers its own receiver and the old instance can be garbage collected.

`./gradlew :analytics-core:test`. New test: after `shutdown()`, a same-`instanceName` rebuild successfully receives a bridged event.

### Checklist
* [ ] Does your PR have a breaking change? No — `EventBridgeContainer.remove()` is `internal`.

---

> [!NOTE]
> **Low Risk**
> Small, internal lifecycle fix with a focused test; no public API change beyond expected shutdown behavior.
> 
> **Overview**
> **Shutdown now drops the per-`instanceName` `EventBridgeContainer` entry** so a new `Amplitude` with the same name can register its own bridge receiver instead of reusing a dead instance’s slot.
> 
> `performShutdown()` calls new internal `EventBridgeContainer.remove(configuration.instanceName)` after timeline/scope teardown. That addresses SDK-171: `EventBridge` only allows one receiver per channel, so without removal, shutdown leaked the old container and cross-SDK / Experiment bridged events could be lost on recreate.
> 
> A test asserts shutdown → rebuild with the same `instanceName` → `sendEvent` on the bridge reaches the new instance’s plugin.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 461a59d4bdd721bec0623105f182c5b28c4a9e21. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/analytics-core/src/main/java/com/amplitude/core/Amplitude.kt
+++ b/analytics-core/src/main/java/com/amplitude/core/Amplitude.kt
@@ -97,6 +97,9 @@ open class Amplitude(
     /** Whether [shutdown] has been invoked. Readable by platform subclasses. */
     protected fun isShutdown(): Boolean = shutdownState.get()
 
+    // The EventBridge container this instance registered on; freed (ownership-checked) on shutdown.
+    @Volatile private var eventBridgeContainer: EventBridgeContainer? = null
+
     @RestrictedAmplitudeFeature
     val diagnosticsClient: DiagnosticsClient by lazy {
         DiagnosticsClientImpl(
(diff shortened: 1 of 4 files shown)


polbins:
bugbot run

PR #423, opened 2026-07-23
fix(core): null-safe filename filter in EventsFileManager (SDK-176)

polbins:
A customer on `analytics-android` v1.21.4 reported an intermittent `NullPointerException` (~0.05% of users) during `flush()` while transferring persisted Identify events. Root cause: `File.listFiles()` can hand our `FilenameFilter` callback a **null** file name on some devices/filesystems. The callback dereferenced `name` without a null check, so Kotlin's non-null parameter intrinsic threw. Tracked as SDK-176.

### Describe the solution
Guard the filter callback against a null name, and collapse the three duplicated `listFiles` predicates in `EventsFileManager` into a single null-safe helper (`listStorageFiles`) plus a testable `internal matchesStorageFile(name: String?, ...)`.

* `./gradlew :analytics-core:test --tests "*EventsFileManagerTest"` — added 4 cases covering the null name and predicate matching.
* `./gradlew :analytics-core:apiCheck` — no public API change (new member is `internal`).

SDK-176

### Checklist
* [ ] Does your PR have a breaking change? No.

---

> [!NOTE]
> **Low Risk**
> Targeted defensive fix in local event file enumeration with no public API change and regression tests; behavior unchanged except avoiding the null-name crash.
> 
> **Overview**
> Fixes an intermittent **NullPointerException** during event flush (SDK-176) when `File.listFiles()` passes a **null** name into the `FilenameFilter` callback on some devices/filesystems.
> 
> **`EventsFileManager`** now routes all directory listing through **`listStorageFiles`** and an **`internal matchesStorageFile(name: String?, …)`** helper that rejects null names before applying the storage-key and predicate checks. **`read()`**, **`currentFile()`**, and **`handleV1Files()`** use this helper instead of three duplicated inline filters.
> 
> Tests add a **`MatchesStorageFile`** nested suite covering null names, storage-key matching, and predicate behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a678061719005c8ab58a47a512d37db2334da8fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/analytics-core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
+++ b/analytics-core/src/main/java/com/amplitude/core/utilities/EventsFileManager.kt
@@ -80,14 +80,29 @@ class EventsFileManager(
     }
 
     /**
+     * Lists this storage's files, applying [matches] to the file name.
+     * Returns an empty array if the directory can't be enumerated.
+     */
+    private fun listStorageFiles(matches: (name: String) -> Boolean): Array<File> =
+        directory.listFiles { _, name -> matchesStorageFile(name, matches) } ?: emptyArray()
+
+    /**
+     * True if [name] belongs to this storage and satisfies [matches].
+     *
+     * [name] is nullable because the platform can hand a null entry to the
+     * [java.io.FilenameFilter] callback on some devices/filesystems (SDK-176).
+     */
+    internal fun matchesStorageFile(
+        name: String?,
+        matches: (name: String) -> Boolean,
+    ): Boolean = name != null && name.contains(storageKey) && matches(name)
+
+    /**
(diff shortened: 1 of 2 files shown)


crleona (approved):
Do we have any idea how we could be passed a null name?

ianmeyer (approved):

polbins:
> Do we have any idea how we could be passed a null name?

My hunch is a directory that was deleted in the middle of an enumeration, other than that it could be a device/OS quirk. Either way the guard is the way to go since it's coming from Java.

PR #432, opened 2026-07-27
fix(android): replace same-name instances automatically (SDK-171)

polbins:

If you create a second Android `Amplitude` with an `instanceName` that is already in use, the first one keeps running. The `Application` still holds the lifecycle callbacks it registered, so it keeps autocapturing, keeps starting sessions, and keeps owning the analytics connector. Creating a new instance with autocapture turned off does not stop the old one (SDK-171).

### Describe the solution

The rule: for one `Application` and instance name, only the newest instance does any work.

- `AmplitudeRegistry` remembers which instance owns each name. A new instance takes the name in `build()`, before its own setup starts, while holding one lock — so if two are created at once exactly one ends up owning the name, and if taking it fails, the instance already there keeps working.
- Taking over sets a flag on the old instance: from then on it will not accept events, start sessions, install plugins, or write to anything shared.
- The old instance immediately stops receiving Activity callbacks, stops accepting events, and stops autocapture. Its remaining plugins are removed once the events it already accepted have been processed, off the caller's thread.

Nothing waits on the old instance, so a stuck setup or a hanging plugin cannot hold up the new one. 

Where to start reading: `AmplitudeRegistry`, then `Amplitude.build` and `Amplitude.retire`, then the checks in `Timeline`. The first commit is a separate fix this one depends on — `WindowCallbackManager` unwrapped any `AutocaptureWindowCallback` it found, so stopping one instance's manager could remove another instance's wrapper.

No public API change and no api dump change.

- `./gradlew ktlintFormat :analytics-core:test :android:test apiCheck`
- `AmplitudeReplacementTest` (13 tests) covers replacing a same-name instance, leaving other names alone, a replaced instance going quiet, replacing one that is still setting up, not waiting for its cleanup, failing to take the name, connector ownership, events already queued, and two instances created at once.

- [SDK-171](https://linear.app/amplitude/issue/SDK-171)

### Checklist
* [x] Does your PR have a breaking change? No — no public API change, `apiCheck` passes. Behaviour does change for apps that relied on two same-name instances both staying active.

---

> [!NOTE]
> **Medium Risk**
> Changes core instance lifecycle and shared-by-name storage, sessions, and connector behavior (no public API change), but behavior breaks apps that depended on multiple same-name instances staying active; coverage is extensive in AmplitudeReplacementTest.
> 
> **Overview**
> Fixes **SDK-171**: creating a second Android `Amplitude` with the same `instanceName` no longer leaves the first instance running autocapture, sessions, and connector ownership.
> 
> **`AmplitudeRegistry`** makes the newest instance per name the sole owner (same `Application` only). Activation happens in `build()` before setup; the previous instance is **retired** without blocking the new one. Retired instances get `isActive == false`, drop new events, skip session side effects while draining the queue, cannot `reset()` shared identity, and bail out of in-flight `buildInternal`. Cleanup unregisters lifecycle callbacks, stops autocapture immediately, closes the timeline, then removes plugins asynchronously after the queue drains.
> 
> **`Timeline`** adds `close()` / `awaitDrained()` and gates `process()` and session handling on active ownership. **Analytics connector** identity and event-bridge setup/writes run only via `AmplitudeRegistry.runIfActive`.
> 
> **`WindowCallbackManager`** splices its wrapper out of a stacked callback chain when another manager wrapped on top, so retiring one instance does not strip another’s autocapture wrapper.
> 
> ProGuard adds `-keepnames` for `UniversalPlugin` implementations for readable retirement logs. **`AmplitudeReplacementTest`** (13 cases) covers takeover, concurrency, failed activation, connector routing, and queued-event drain.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 271b195882c08873c13ad1168acb38748dae8ca8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -14,6 +14,7 @@ import com.amplitude.android.utilities.ActivityLifecycleObserver
 import com.amplitude.core.RestrictedAmplitudeFeature
 import com.amplitude.core.State
 import com.amplitude.core.diagnostics.DiagnosticsContextProvider
+import com.amplitude.core.platform.UniversalPlugin
 import com.amplitude.core.platform.plugins.AmplitudeDestination
 import com.amplitude.core.platform.plugins.ContextPlugin
 import com.amplitude.core.platform.plugins.GetAmpliExtrasPlugin
@@ -24,8 +25,10 @@ import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.launch
 import java.io.File
 import java.lang.ref.WeakReference
+import java.util.concurrent.atomic.AtomicBoolean
 import com.amplitude.core.Amplitude as CoreAmplitude
 
 @OptIn(ExperimentalCoroutinesApi::class, RestrictedAmplitudeFeature::class)
@@ -58,6 +61,17 @@ open class Amplitude internal constructor(
     internal lateinit var autocaptureManager: AutocaptureManager
         private set
 
+    // Assigned in [build], not by a field initializer: initializers run after the core constructor,
+    // where they would clobber a retirement from a same-name instance racing this one.
+    private lateinit var retired: AtomicBoolean
+
+    /** Whether this instance still owns its instance name. Gates all work an active instance does. */
+    internal val isActive: Boolean
+        get() = !retired.get()
+
+    internal val application: Application
+        get() = (configuration as Configuration).context as Application
+
     /**
      * Init-order trap: [CoreAmplitude]'s `init` calls [build] before this class's field
      * initializers run, and [buildInternal] may already be on a background thread.
@@ -65,6 +79,7 @@ open class Amplitude internal constructor(
      * initializers, or `init {}`.
      */
     override fun build(): Deferred<Boolean> {
+        retired = AtomicBoolean(false)
         activityLifecycleCallbacks = ActivityLifecycleObserver()
         androidContextPlugin = AndroidContextPlugin()
         val androidConfig = configuration as Configuration
@@ -76,15 +91,10 @@ open class Amplitude internal constructor(
                 logger = logger,
                 diagnosticsClient = diagnosticsClient,
             )
+        // Claim the instance name before the build exists, so nothing the build installs can run
+        // before the claim, and a failed claim leaves no build behind.
+        AmplitudeRegistry.activate(this)
         return super.build()
-    }
-
-    init {
-        registerShutdownHook()
-
-        with(configuration.context as Application) {
-            registerActivityLifecycleCallbacks(activityLifecycleCallbacks)
-        }
     }
 
     override fun createTimeline(): Timeline {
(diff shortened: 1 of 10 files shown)


polbins:
bugbot run

polbins:
bugbot run

android/src/main/java/com/amplitude/android/Timeline.kt
```diff
@@ -99,27 +121,36 @@ class Timeline(
...
 
             END_SESSION_EVENT -> {
                 // No specific action needed before processing this event type
             }
 
             else -> {
-                if (event.sessionId != DEFAULT_SESSION_ID) {
+                if (event.sessionId != DEFAULT_SESSION_ID && ownsSessions()) {
                     val stopAndStartSessionEvents = startNewSessionIfNeeded(eventTimestamp)
                     processAndPersistEvents(stopAndStartSessionEvents)
```
ianmeyer, line 155: Should this path call `processAndPersistEvents` twice? It falls through to `processAndPersistEvents(listOf(event))` below.

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -159,6 +180,42 @@ open class Amplitude internal constructor(
...
+
+        // The rest go last: they must outlive an in-flight build that is still adding them and the
+        // queued events still draining through them. Off the constructor's thread either way, so
+        // the replacement never waits on this.
+        amplitudeScope.launch(amplitudeDispatcher) {
+            isBuilt.join()
+            (timeline as Timeline).awaitDrained()
+            plugins(UniversalPlugin::class.java).forEach { plugin ->
+                runCatching { remove(plugin) }
+                    .onFailure { logger.warn("Failed to remove ${plugin.name ?: plugin::class.java.simpleName}: $it") }
```
ianmeyer, line 217: Is this logging ever used in production obfuscated code? If so, we'd want to add a `Plugin.name` so it isn't obfuscated.

ianmeyer (approved):
I did my best to follow along

android/src/main/java/com/amplitude/android/Timeline.kt
polbins: Good question! The first call is for session events (`session_end`/`session_start`), that ensures we've ended/started a new session if needed before we process the actual event (fall-through case).

android/src/main/java/com/amplitude/android/Amplitude.kt
polbins: It is going to be on production, adding this to our rules!

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -134,6 +152,9 @@ open class Amplitude internal constructor(
     }
 
     override fun reset(): Amplitude {
+        // Identity is shared by instance name, so a replaced instance must not rotate it.
+        if (!isActive) return this
```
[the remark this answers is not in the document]
polbins: This is a case of misuse and predates this PR. We have `reset()` gated because it's open and would likely be called on teardown, but `setUserId`/`setDeviceId` are final in core. Will leave this as-is.

android/src/main/java/com/amplitude/android/Timeline.kt
```diff
@@ -99,27 +121,36 @@ class Timeline(
...
+                }
             }
 
             END_SESSION_EVENT -> {
                 // No specific action needed before processing this event type
             }
 
             else -> {
-                if (event.sessionId != DEFAULT_SESSION_ID) {
+                if (event.sessionId != DEFAULT_SESSION_ID && ownsSessions()) {
```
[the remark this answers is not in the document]
polbins: `lastEventId` is an in-memory copy of a value that lives in storage shared by instance name, so two same-name timelines have always been able to hand out the same ids. This predates the PR, where here it's bounded to the drain. `event_id` also isn't what dedupes server-side. Will leave this as-is.

PR #433, opened 2026-07-28
fix(android): stop network callbacks leaking Amplitude instances (SDKA-31)

### LINKED ISSUE #236: AndroidNetworkListener.startListening() should not ignore TooManyRequestsException
## Expected Behavior
Surface the crash to allow the developer to debug closer to the source of the issue as Google intended.

## Current Behavior
AndroidNetworkListener.startListening() currently swallows the exception.

```
    fun startListening() {
        try {
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
                setupNetworkCallback() // <- this calls registerNetworkCallback
            } else {
                setupBroadcastReceiver()
            }
        } catch (throwable: Throwable) {
            // We've seen issues where we see exceptions being thrown by connectivity manager
            // which crashes an app. Its safe to ignore these exceptions since we try our best
            // to mark a device as offline
            // Github Issues:
            // https://github.com/amplitude/Amplitude-Kotlin/issues/220
            // https://github.com/amplitude/Amplitude-Kotlin/issues/197
            logger.warn("Error starting network listener: ${throwable.message}")
        }
    }
```

The issues linked don't show understanding or research into the root cause. Whether its mishandling of Amplitude sdk or another sdk leaking the NetworkCallback calls.

## Possible Solution / Reasoning
Rethrow the crash immediately, and perhaps link to and update documentation to help developers find the root cause. Point out tear down call when showing how to configure Amplitude sdk https://amplitude.com/docs/sdks/analytics/android/android-kotlin-sdk#configure-the-sdk

Perhaps encourage a singleton instance of the Amplitude class, and detail how it can leak ConnectivityManager$NetworkCallback and lead to TooManyRequestsException if multiple instances are created. That's more your expertise.

Reasoning:
https://developer.android.com/reference/android/net/ConnectivityManager#registerNetworkCallback(android.net.NetworkRequest,%20android.net.ConnectivityManager.NetworkCallback)
ConnectivityManager.requestNetworkCallback (and a few other entry points) has the following documentation:

```
To avoid performance issues due to apps leaking callbacks, the system will limit the number of outstanding requests to 100 per app (identified by their UID), shared with all variants of this method, of requestNetwork as well as ConnectivityDiagnosticsManager. registerConnectivityDiagnosticsCallback. Requesting a network with this method will count toward this limit. If this limit is exceeded, an exception will be thrown. To avoid hitting this issue and to conserve resources, make sure to unregister the callbacks with unregisterNetworkCallback(ConnectivityManager. NetworkCallback).
```

Meaning if a library is leaking its `NetworkCallback` the 100 concurrent callbacks registered can be hit fairly quickly. By swallowing the exception it is making it more difficult for developers to debug and find the root cause.

For us we were creating a new instance of the `Amplitude` class for every `track` request and not tearing it down properly. This was only caught as I was looking for `NetworkCallback` in the memory dump and saw nearly a hundred instances of `networkCallbackForHigherApiLevels` in memory. Genuinely well done on the naming though, I wouldn't have caught it if it was named badly.

I've seen WorkManager cop a lot of blame for leaking due to libraries sometimes applying this bandaid fix for a bug that should be surfaced. I feel as Android devs we should be surfacing this exception immediately, and then offering help to developers to track down the root cause.

Thanks for your time!

## Steps to Reproduce
1. Create ~100 instances of the Amplitude sdk without tearing down. (Misuse I agree, but hard to track and could be helped with easy documentation updates for future players)
2. 
3.
4.

## Environment
- Unity Plugin Version: 1.19.4 `amplitude = { group = "com.amplitude", name = "analytics-android", version.ref = "amplitude" }`
- Device: reproduced Across Pixel and Samsung
- Device OS and Version: reproduced in at least 14 and 15

ianmeyer:

Fixes #236. The `ConnectivityManager.NetworkCallback` registered by `AndroidNetworkListener` had two problems:

1. **Memory leak** — the callback strongly referenced the `NetworkChangeCallback`, which holds the entire `Amplitude` instance. Since `ConnectivityManager` keeps registered callbacks strongly reachable until unregistered, an abandoned instance could never be garbage collected (the one system registration missed by #425).
2. **Callback quota exhaustion** — `unregisterNetworkCallback()` was never called in practice: plugin `teardown()` only runs on explicit `remove(plugin)`, which apps don't call. Each abandoned instance permanently consumed one of the app's 100 network-callback slots, eventually causing `TooManyRequestsException` for the host app and other SDKs.

Linear: [SDKA-31](https://linear.app/amplitude/issue/SDKA-31/kotlin-sdk-connectivitymanager-network-callback-leaks-amplitude)

### Describe the solution

* Extracted the anonymous callback into `AndroidNetworkCallback`, which holds the delegate through a `WeakReference` — the system side can no longer pin an Amplitude instance.
* A callback whose delegate has been collected frees its registration slot two ways: it self-unregisters on the next network event it receives, and every new `startListening()` sweeps a registry of live registrations (covers the "new instance per `track()`" misuse from the issue without waiting for a network event).
* `stopListening()` remains the deterministic path and is now idempotent; `startListening()` guards double registration.
* `AndroidNetworkState` (renamed from the nested `NetworkState`) shares the same `WeakReference` and notifies implicitly on creation and on toggles; each class now lives in its own file.
* Kept the catch-all in `startListening()` rather than rethrowing `TooManyRequestsException`: crashing host apps from an analytics SDK is worse than degraded offline detection, and with this fix the SDK no longer fills the quota.

No public API changes (`apiCheck` passes; new classes are `internal`).

* `./gradlew :android:test :android:apiCheck :android:ktlintCheck` — all green.
* New unit tests: `AndroidNetworkCallbackTest` and `AndroidNetworkStateTest` cover the bridge and the state machine; `AndroidNetworkListenerTest` adds GC-based end-to-end tests proving an abandoned listener's delegate is collectable and its registration is released, plus sweep selectivity (live callbacks are spared), stop idempotency, and re-setup after stop.
* Key regressions were mutation-tested: dropping the sweep's collected-delegate guard or the `stopListening()` null-out fails the new tests.

* https://github.com/amplitude/Amplitude-Kotlin/issues/236
* [`ConnectivityManager.registerNetworkCallback` docs](https://developer.android.com/reference/android/net/ConnectivityManager#registerNetworkCallback(android.net.NetworkRequest,%20android.net.ConnectivityManager.NetworkCallback)) — 100-callback limit
* Prior art: #425 (same weak-reference pattern for other system-held references)

### Checklist

<details>
<summary>AI Prompt</summary>

From GitHub issue #236, two issues to solve while keeping the architecture of AndroidNetworkListener clean: (1) a memory leak in AndroidNetworkListener.setupNetworkCallback() — the callback passed to registerNetworkCallback leaks the entire Amplitude instance; (2) ConnectivityManager.unregisterNetworkCallback() is never called, so if the Amplitude instance is released and recreated, the ConnectivityManager approaches its quota of 100 simultaneous callbacks.

</details>

---

> [!NOTE]
> **Medium Risk**
> Touches offline detection and ConnectivityManager registration lifecycle; behavior is intended to match prior logic but incorrect cleanup could miss network events or unregister live callbacks.
> 
> **Overview**
> Fixes **memory leaks** and **`ConnectivityManager`’s 100-callback limit** when apps abandon `Amplitude` without tearing down the network listener.
> 
> Network handling is split into **`AndroidNetworkCallback`** and **`AndroidNetworkState`**, both holding the app delegate via **`WeakReference`** so a registered system callback no longer pins the `Amplitude` instance. When the delegate is collected, the callback **unregisters itself** on the next network event; **`unregisterAbandonedCallbacks()`** runs on each new **`setupNetworkCallback()`** so slots are reclaimed without waiting for connectivity changes.
> 
> **`AndroidNetworkListener`** now delegates registration to **`AndroidNetworkCallback`**, skips duplicate setup, and makes **`stopListening()`** idempotent by clearing the registered callback before unregistering. Behavior for live listeners is unchanged; new unit tests cover GC abandonment, sweep selectivity, and stop/setup lifecycle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ade49b9cff5b0ea9f8a70ca4597c2113a4630705. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/android/src/main/java/com/amplitude/android/utilities/AndroidNetworkCallback.kt
+++ b/android/src/main/java/com/amplitude/android/utilities/AndroidNetworkCallback.kt
@@ -1,0 +1,136 @@
+package com.amplitude.android.utilities
+
+import android.Manifest.permission.ACCESS_NETWORK_STATE
+import android.net.ConnectivityManager
+import android.net.ConnectivityManager.NetworkCallback
+import android.net.Network
+import android.net.NetworkCapabilities
+import android.net.NetworkCapabilities.NET_CAPABILITY_INTERNET
+import android.net.NetworkCapabilities.NET_CAPABILITY_VALIDATED
+import android.net.NetworkRequest
+import android.os.Build.VERSION
+import android.os.Build.VERSION_CODES
+import androidx.annotation.RequiresPermission
+import java.lang.ref.WeakReference
+import java.util.concurrent.CopyOnWriteArrayList
+
+/**
+ * Bridges [ConnectivityManager] network callbacks to a [AndroidNetworkListener.NetworkChangeCallback]
+ * held through a [WeakReference]. The system keeps registered callbacks strongly reachable, so a
+ * strong reference here would pin the delegate — and the Amplitude instance behind it — after the
+ * host app drops the instance without calling [AndroidNetworkListener.stopListening].
+ *
+ * The system also limits an app to 100 outstanding callbacks, so a callback whose delegate has
+ * been garbage collected frees its registration slot: on the next network event it receives, or
+ * when any new [AndroidNetworkListener] starts listening.
+ */
+internal class AndroidNetworkCallback(
+    delegate: AndroidNetworkListener.NetworkChangeCallback,
+    private val connectivityManager: ConnectivityManager,
+) : NetworkCallback() {
+    private val delegateRef = WeakReference(delegate)
+    private var _networkState: AndroidNetworkState? = null
+
+    /**
+     * Returns the tracked network state, or null if no network is tracked yet or the delegate
+     * was garbage collected (in which case this callback's registration is released).
+     */
+    private val networkState: AndroidNetworkState?
+        get() = networkChangeCallback?.let { _networkState }
+
+    /**
+     * The delegate, or null — after releasing this callback's registration — if the delegate,
+     * along with the Amplitude instance that owned it, was garbage collected.
+     */
+    private val networkChangeCallback: AndroidNetworkListener.NetworkChangeCallback?
+        get() =
+            delegateRef.get() ?: run {
+                safeUnregister()
+                null
+            }
+
+    /** Registers this callback with [ConnectivityManager] and starts tracking it. */
+    fun register(networkRequest: NetworkRequest) {
+        connectivityManager.registerNetworkCallback(networkRequest, this)
+        registeredCallbacks.add(this)
+    }
+
+    @RequiresPermission(ACCESS_NETWORK_STATE)
+    override fun onAvailable(network: Network) {
+        networkChangeCallback ?: return
+
+        // A default network is available; the new state notifies the delegate on creation
+        val capabilities = connectivityManager.getNetworkCapabilities(network)
+        _networkState =
+            AndroidNetworkState(
+                network = network,
+                delegateRef = delegateRef,
+                available = capabilities?.available() ?: true,
+                blocked = false,
+            )
+    }
+
+    override fun onUnavailable() {
+        // no network is found
+        networkChangeCallback?.onNetworkUnavailable()
+    }
+
+    override fun onLost(network: Network) {
+        networkState?.update(network, available = false)
+    }
+
+    override fun onCapabilitiesChanged(
+        network: Network,
+        networkCapabilities: NetworkCapabilities,
+    ) {
+        networkState?.update(network, available = networkCapabilities.available())
+    }
+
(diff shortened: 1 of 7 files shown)


ianmeyer:
bugbot run

polbins (approved):
LGTM! Thanks for the continued cleanup! 🙏🏽

PR #446, opened 2026-07-31
feat(core): new upload retry backoff schedule (SDKA-36)

ianmeyer:

Upload retries used a `1s → 2s → 4s → 8s → 16s` exponential formula, then paused for a 60s cooldown once `flushMaxRetries` was hit and restarted the ladder. Every client retried on the same unjittered cadence, and a 429 was treated like any other retryable failure, so throttled clients came back well before the server wanted them to.

The [Client SDK Retry Logic](https://linear.app/amplitude/document/client-sdk-retry-logic-69d8c83f50b2) doc defines one retry policy across the Swift, Kotlin, and TypeScript SDKs. This aligns Kotlin with it.

Ref: SDKA-36

### Describe the solution

`ExponentialBackoffRetryHandler` now walks a fixed schedule — 1s, 8s, 32s, 128s — and then repeats 300s for as long as uploads keep failing. Each delay is jittered by ± half of itself, so clients spread out instead of retrying in lockstep; the effective windows are 0.5–1.5s, 4–12s, 16–48s, 64–192s, then 150–450s. A successful upload calls `reset()` and the schedule starts over.

Three deliberate decisions worth reviewing:

* **Retries are no longer capped.** The old max-attempt cooldown restarted the ladder back at the shortest delay, which contradicts "300s repeats until success." Removing the cap means `configuration.flushMaxRetries` no longer governs upload retries — it still applies to per-event attempts in `InMemoryResponseHandler`. That is a behavior change for anyone who set it expecting a ceiling. Events are not dropped by this path, so nothing is lost; uploads just keep trying.
* **A 429 waits a fixed 30s** and bypasses the backoff schedule entirely, per the doc's "429 (wait 30 sec)". Note this delay is not jittered and does not escalate, so sustained throttling produces a flat 30s cadence across clients. Respecting `Retry-After` is not possible yet — `HttpClientInterface.upload()` returns no headers.
* **An invalid API key is terminal.** The response handler still drops the batch and fires the 400 callbacks; the pipeline now stops walking the remaining event files instead of re-issuing a request per file that is guaranteed to fail the same way.

`ExponentialBackoffRetryHandler` is annotated `@RestrictedAmplitudeFeature`, so external construction now requires opt-in. The public API dump is unchanged — `apiCheck` passes with no `.api` edits. The old `(Int, Int, Double)` constructor and `maxRetryAttempt` are retained but deprecated and ignored.

* `./gradlew :analytics-core:test :android:test apiCheck ktlintCheck` — green.
* `ExponentialBackoffRetryHandlerTest` asserts the ladder in order, that it plateaus at 300s across 10 further attempts, that every delay lands within ± half of its nominal value and that the values actually vary, and that `reset()` returns to the 1s step. Delays are measured in `runTest` virtual time, so the suite does not sleep.
* `EventPipelineTest` covers the 429 path waiting exactly 30s without touching the retry handler, an invalid API key not retrying, and retries continuing past `flushMaxRetries` (attempt 50 → 51).

* [Client SDK Retry Logic](https://linear.app/amplitude/document/client-sdk-retry-logic-69d8c83f50b2)
* [SDKA-36](https://linear.app/amplitude/issue/SDKA-36/align-upload-retry-backoff-with-the-client-sdk-retry-logic-doc)

### Checklist

Not marked as breaking, so semantic-release cuts a minor. Arguable both ways: `flushMaxRetries` becomes inert for uploads, and the deprecated constructor lost its parameter defaults (a source break, though binary compatibility holds). Say the word and I will add a `BREAKING CHANGE:` footer for a major.

<details>
<summary>AI Prompt</summary>

Update retry logic in ExponentialBackoffRetryHandler, etc.

New logic:

Retry exponentially with delays 8s, 32s, 128s, 256s, 300s, 300s, 300s. For each stage add a jitter of +/- half the delay.

so, 4-12s, 16-48s, 64-192s, 128s-384s, 150s-450s, 150s-450s, ...

do not retry on bad API key errors, this is terminal.

for 429 errors, wait a minimum of 30 seconds before retrying.

(The schedule was subsequently changed to 1s, 8s, 32s, 128s, 300s to match the design doc, and the 429 delay became a fixed 30s rather than a minimum.)

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes core upload retry timing and removes the upload retry ceiling (`flushMaxRetries` no longer applies here), which can alter offline recovery behavior and server load patterns for existing apps.
> 
> **Overview**
> Aligns Kotlin **event upload retries** with the shared Client SDK retry policy: fixed backoff steps with jitter, special handling for **429** and **invalid API key**, and **no cap** on upload retries.
> 
> **`ExponentialBackoffRetryHandler`** no longer uses exponential `2^n` delays or `flushMaxRetries`. Retries follow **1s → 8s → 32s → 128s**, then **300s** indefinitely, each step jittered by ±50%. `attemptRetry` always signals retry (`canRetry` is always true). The old parameterized constructor and `maxRetryAttempt` / `maxDelayInMs` are deprecated. The class is **`@RestrictedAmplitudeFeature`**.
> 
> **`EventPipeline`** wires the new handler (default constructor only), drops the **max-retry cooldown** (`MAX_RETRY_ATTEMPT_SIG` / 60s pause), treats **invalid API key** `BadRequestResponse` as **terminal** (stop processing remaining files, no retry), and on **429** waits a **fixed 30s** before re-queueing upload—**without** going through the backoff handler.
> 
> Tests cover the new schedule, jitter, 429 timing, invalid API key, and retries continuing past former `flushMaxRetries` limits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 260e9ddd0f91ad9f1ec20cbd1bffb488ed8e42fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/analytics-core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
+++ b/analytics-core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
@@ -185,7 +185,7 @@ public class EventPipeline(
         scope.launch(amplitude.storageIODispatcher) {
             if (isActive && running && !scheduled) {
                 scheduled = true
-                delay(amplitude.configuration.flushIntervalMillis.toLong())
+                delay(amplitude.configuration.flushIntervalMillis.milliseconds)
                 flush()
                 scheduled = false
             }
--- a/analytics-core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
+++ b/analytics-core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
@@ -1,12 +1,15 @@
 package com.amplitude.core.platform
 
 import com.amplitude.core.Amplitude
+import com.amplitude.core.RestrictedAmplitudeFeature
 import com.amplitude.core.Storage
 import com.amplitude.core.events.BaseEvent
 import com.amplitude.core.utilities.ExponentialBackoffRetryHandler
+import com.amplitude.core.utilities.http.BadRequestResponse
 import com.amplitude.core.utilities.http.HttpClient
 import com.amplitude.core.utilities.http.HttpClientInterface
 import com.amplitude.core.utilities.http.ResponseHandler
+import com.amplitude.core.utilities.http.TooManyRequestsResponse
 import com.amplitude.core.utilities.logWithStackTrace
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.channels.Channel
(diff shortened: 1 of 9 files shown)


analytics-core/src/main/java/com/amplitude/core/platform/EventPipeline.kt
polbins: nit: warning fix
```suggestion
                delay(amplitude.configuration.flushIntervalMillis.milliseconds)
```

polbins (commented):
Should we deprecate `flushMaxRetries` as well? (this might be a question for the wider team)

ianmeyer:
@polbins deprecating `flushMaxRetries` seems correct to me

polbins (approved):

PR #469, opened 2026-08-13
fix: safe-parse non-JSON 4xx upload responses

polbins:
Custom `serverUrl` proxies that return non-JSON 400/413/429 caused `AnalyticsResponse.create()` to throw before `responseHandler.handle()` ran. The event file was never released, removed, or split — permanent silent stall ([SDK-215](https://linear.app/amplitude/issue/SDK-215/android-sdk-strands-events-on-non-json-4xx-proxy-response)).

### Describe the solution
Route 400/413/429 through `parseResponseBodyOrGetDefault()` (same as 5xx). Keep status-specific types so existing handlers still run (proxy 400 → release+retry, 413 → split, 429 → retry).

Intent: stop stranding and match Swift/TS parse resilience. This does **not** turn a persistent plain-text 400 into a successful upload — proxies should still return [HTTP V2 JSON](https://amplitude.com/docs/apis/analytics/http-v2#response).

* `./gradlew :analytics-core:test --tests "com.amplitude.core.utilities.http.AnalyticsResponseTest" --tests "com.amplitude.core.utilities.HttpClientTest" --tests "com.amplitude.core.utilities.FileResponseHandlerTest"`
* Plain-text/null 400/413/429 parse without throw; plain-text 400 releases; plain-text 413 splits.

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes upload error parsing and invalid-API-key detection for malformed 4xx bodies; behavior shifts from discarding/stranding to retry/split, which is the intended fix but affects the event pipeline on proxy errors.
> 
> **Overview**
> Fixes **silent event-file stalls** when a custom `serverUrl` proxy returns plain-text or non-JSON bodies for HTTP 400, 413, or 429. Those statuses now go through `parseResponseBodyOrGetDefault()` (same path as 5xx) so `AnalyticsResponse.create()` no longer throws before upload response handling runs.
> 
> **Unparseable bodies** no longer copy the raw text into the JSON `error` field; they yield an empty `JSONObject` instead. That keeps proxy strings like `invalid_api_key` from tripping **`isInvalidApiKeyResponse()`** and dropping batches as a terminal invalid-key case. Status-specific types are unchanged, so handlers still run: proxy 400 → release and retry, 413 → split, 429 → retry.
> 
> Tests cover parsing, `HttpClient` uploads, and `FileResponseHandler` behavior for plain-text 4xx bodies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46400b8084299231c916580ba96c6405b3507a6b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/analytics-core/src/main/java/com/amplitude/core/utilities/http/AnalyticsResponse.kt
+++ b/analytics-core/src/main/java/com/amplitude/core/utilities/http/AnalyticsResponse.kt
@@ -20,9 +20,9 @@ public sealed class AnalyticsResponse(public val status: HttpStatus) {
         ): AnalyticsResponse {
             return when (responseCode) {
                 in SUCCESS.range -> SuccessResponse()
-                in BAD_REQUEST.range -> BadRequestResponse(JSONObject(responseBody))
-                in PAYLOAD_TOO_LARGE.range -> PayloadTooLargeResponse(JSONObject(responseBody))
-                in TOO_MANY_REQUESTS.range -> TooManyRequestsResponse(JSONObject(responseBody))
+                in BAD_REQUEST.range -> BadRequestResponse(parseResponseBodyOrGetDefault(responseBody))
+                in PAYLOAD_TOO_LARGE.range -> PayloadTooLargeResponse(parseResponseBodyOrGetDefault(responseBody))
+                in TOO_MANY_REQUESTS.range -> TooManyRequestsResponse(parseResponseBodyOrGetDefault(responseBody))
                 in TIMEOUT.range -> TimeoutResponse()
                 else -> FailedResponse(parseResponseBodyOrGetDefault(responseBody))
             }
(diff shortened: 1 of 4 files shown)


analytics-core/src/main/java/com/amplitude/core/utilities/http/AnalyticsResponse.kt
crleona: Should we return this as a fake error response? Looking specifically at `isInvalidApiKeyResponse`, this just checks if "invalid api key" is in the response, and we may not want to honor that from anything but a well formed response.

crleona (approved):

analytics-core/src/main/java/com/amplitude/core/utilities/http/AnalyticsResponse.kt
polbins: Yeah, I think that would be safer. 👍🏽

PR #471, opened 2026-08-20
feat: capture uncaught Android exceptions for diagnostics

ianmeyer:
Android apps can crash without the SDK reporting anything useful through diagnostics. This adds capture of uncaught Java/Kotlin exceptions and reports a prior-session crash through the existing diagnostics pipeline, matching the iOS schema (`analytics.crash` counter + event with a `report` property).

### Describe the solution
- **`CrashCatcher` / `CrashStorage`**: install an uncaught-exception handler that streams a bounded crash report to disk (64 frames / 4 cause levels), chains to the previous handler, and kills the process when there is no handler to chain to.
- **Process-wide dedupe**: persist each throwable once when multiple Amplitude handlers are chained (stacked instances or wrapped handlers).
- **`Amplitude` init**: consume the prior crash on launch and record it via an internal `DiagnosticsClient.recordCrash` extension in `:android` (no public API change).
- **Sampling**: buffer the crash counter/event like other diagnostics data; upload remains gated on the session being sampled in.
- **Sample app**: add a Compose **Crash** button for manual end-to-end testing.

- [x] `./gradlew test ktlintCheck apiCheck`
- [x] Run the sample app, open Compose, tap **Crash**, relaunch, and confirm diagnostics receives an `analytics.crash` event when the session is sampled in
- [x] Unit tests: `CrashCatcherTest`, `CrashStorageTest`, `CrashDiagnosticsTest`

SDKA-65

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Installs a process-wide uncaught-exception handler and synchronous crash I/O on the crash path, with lifecycle coupling to instance retirement and diagnostics sampling—important for correctness but not end-user auth or payment flows.
> 
> **Overview**
> Adds **Android crash diagnostics** aligned with iOS: uncaught exceptions that involve Amplitude code can be persisted and reported on the next launch as `analytics.crash` (counter + event with a `report` property).
> 
> **Core hook:** `Amplitude` now runs config validation and a new protected **`initWatchers()`** in the first `init` block, before timeline/build, so process-wide watchers can cover early init failures. **`DiagnosticsClient`** exposes **`shouldTrack`** so crash persistence can respect diagnostics sampling.
> 
> **Android:** **`CrashCatcher`** chains the default uncaught-exception handler, writes a bounded stack report to disk when remote config enables crash tracking and the session is sampled in, then on **`buildInternal`** the SDK consumes that file and calls an internal **`recordCrash`** extension. **`CrashTrackingRemoteConfig`** gates feature availability by SDK semver from diagnostics remote config. **`detachWatchers()`** runs when build/activation fails or an instance retires so half-built instances do not keep persisting crashes (the handler stays installed but releases its references).
> 
> The Kotlin sample app adds a **Crash** button for manual testing; tests cover handler chaining, dedupe, detach paths, and storage bounds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3f864907ed47300546258a361fdd10533b44c82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/analytics-core/src/main/java/com/amplitude/core/diagnostics/DiagnosticsClientImpl.kt
+++ b/analytics-core/src/main/java/com/amplitude/core/diagnostics/DiagnosticsClientImpl.kt
@@ -46,8 +46,9 @@ internal class DiagnosticsClientImpl(
     private val startTimestampSeed: String = startTimestamp.toString()
     private var sampleRate: Double = sampleRate.coerceIn(0.0, 1.0)
 
-    private var shouldTrack: Boolean =
+    override var shouldTrack: Boolean =
         enabled && Sample.isInSample(seed = startTimestampSeed, sampleRate = this.sampleRate)
+        private set
 
     private var didSetBasicTags: Boolean = false
     private var didFlushPreviousSession: Boolean = false
--- a/analytics-core/src/main/java/com/amplitude/core/diagnostics/DiagnosticsClientImpl.kt
+++ b/analytics-core/src/main/java/com/amplitude/core/diagnostics/DiagnosticsClientImpl.kt
@@ -46,6 +46,7 @@ internal class DiagnosticsClientImpl(
     private val startTimestampSeed: String = startTimestamp.toString()
     private var sampleRate: Double = sampleRate.coerceIn(0.0, 1.0)
 
+    @Volatile
     override var shouldTrack: Boolean =
         enabled && Sample.isInSample(seed = startTimestampSeed, sampleRate = this.sampleRate)
         private set
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -2,6 +2,8 @@ package com.amplitude.android
 
 import android.app.Application
 import android.content.Context
+import com.amplitude.android.crash.CrashCatcher
+import com.amplitude.android.crash.recordCrash
 import com.amplitude.android.diagnostics.AndroidDiagnosticsContextProvider
 import com.amplitude.android.migration.MigrationManager
 import com.amplitude.android.plugins.AnalyticsConnectorIdentityPlugin
@@ -48,6 +50,18 @@ public open class Amplitude internal constructor(
         storageIODispatcher = storageIODispatcher,
     ) {
     public constructor(configuration: Configuration) : this(configuration, State())
+
+    // Assigned in [initWatchers], after configuration validation and before any other
+    // initialization, so the handler covers SDK init without leaking on invalid config.
+    private lateinit var crashCatcher: CrashCatcher
+
+    override fun initWatchers() {
+        crashCatcher =
+            CrashCatcher(
+                context = (configuration as Configuration).context,
+                ioDispatcher = storageIODispatcher,
+            )
+    }
 
     override val sessionId: Long
         get() {
@@ -121,6 +135,10 @@ public open class Amplitude internal constructor(
         // A build that outlives its instance must not install plugins or claim shared state:
         // the replacement owns both by then.
         if (!isActive) return
+
+        crashCatcher.consumePreviousCrash()?.let { previousCrash ->
+            diagnosticsClient.recordCrash(previousCrash)
+        }
 
         val migrationManager = MigrationManager(this)
         migrationManager.migrateOldStorage()
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -60,6 +60,7 @@ public open class Amplitude internal constructor(
             CrashCatcher(
                 context = (configuration as Configuration).context,
                 ioDispatcher = storageIODispatcher,
+                diagnosticsClientLazy = lazy { diagnosticsClient },
             )
     }
 
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -104,24 +104,32 @@ public open class Amplitude internal constructor(
         retired = AtomicBoolean(false)
         activityLifecycleCallbacks = ActivityLifecycleObserver()
         androidContextPlugin = AndroidContextPlugin()
-        val androidConfig = configuration as Configuration
-        crashTrackingRemoteConfig =
-            CrashTrackingRemoteConfig(
-                remoteConfigClient = remoteConfigClient,
-                sdkVersion = BuildConfig.AMPLITUDE_VERSION,
-            )
-        crashTrackingRemoteConfig.initialize()
-        autocaptureManager =
-            AutocaptureManager(
-                initialAutocapture = androidConfig.autocapture,
-                initialInteractionsOptions = androidConfig.interactionsOptions,
-                remoteConfigClient = if (androidConfig.enableAutocaptureRemoteConfig) remoteConfigClient else null,
-                logger = logger,
-                diagnosticsClient = diagnosticsClient,
-            )
-        // Claim the instance name before the build exists, so nothing the build installs can run
-        // before the claim, and a failed claim leaves no build behind.
-        AmplitudeRegistry.activate(this)
+        try {
+            val androidConfig = configuration as Configuration
+            crashTrackingRemoteConfig =
+                CrashTrackingRemoteConfig(
+                    remoteConfigClient = remoteConfigClient,
+                    sdkVersion = BuildConfig.AMPLITUDE_VERSION,
+                )
+            crashTrackingRemoteConfig.initialize()
+            autocaptureManager =
+                AutocaptureManager(
+                    initialAutocapture = androidConfig.autocapture,
+                    initialInteractionsOptions = androidConfig.interactionsOptions,
+                    remoteConfigClient = if (androidConfig.enableAutocaptureRemoteConfig) remoteConfigClient else null,
+                    logger = logger,
+                    diagnosticsClient = diagnosticsClient,
+                )
+            // Claim the instance name before the build exists, so nothing the build installs can run
+            // before the claim, and a failed claim leaves no build behind.
+            AmplitudeRegistry.activate(this)
+        } catch (error: Throwable) {
+            // Watchers are installed in the constructor, before this method. A throw here never
+            // reaches retire(), so drop them before the half-built instance is abandoned.
+            markRetired()
+            detachWatchers()
+            throw error
+        }
         return super.build()
     }
 
@@ -148,7 +156,9 @@ public open class Amplitude internal constructor(
         if (!isActive) return
 
         crashCatcher.consumePreviousCrash()?.let { previousCrash ->
-            diagnosticsClient.recordCrash(previousCrash)
+            AmplitudeRegistry.runIfActive(this) {
+                diagnosticsClient.recordCrash(previousCrash)
+            }
         }
 
         val migrationManager = MigrationManager(this)
@@ -222,6 +232,17 @@ public open class Amplitude internal constructor(
     internal fun markRetired(): Boolean = retired.compareAndSet(false, true)
 
     /**
+     * Releases process-wide watchers started in [initWatchers]. Safe before [retire], including
+     * when construction fails after those watchers are installed.
+     */
+    internal fun detachWatchers() {
+        if (::crashCatcher.isInitialized) {
+            runCatching { crashCatcher.detach() }
+                .onFailure { logger.warn("Failed to detach the crash handler: $it") }
+        }
+    }
+
+    /**
      * Releases what an inactive instance must not keep holding. Nothing here waits on this
      * instance, and queued events still drain to the storage the replacement shares by name.
      */
@@ -234,6 +255,9 @@ public open class Amplitude internal constructor(
         // takes no part in draining, so it goes now rather than behind the drain below.
         runCatching { findPlugin<AndroidLifecyclePlugin>()?.let { remove(it) } }
             .onFailure { logger.warn("Failed to stop Android autocapture: $it") }
+        // The uncaught exception handler cannot be unregistered, so it has to let go of this
+        // instance instead. Crash persistence belongs to whichever instance is still active.
+        detachWatchers()
 
         // The rest go last: they must outlive an in-flight build that is still adding them and the
         // queued events still draining through them. Off the constructor's thread either way, so
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -3,6 +3,7 @@ package com.amplitude.android
 import android.app.Application
 import android.content.Context
 import com.amplitude.android.crash.CrashCatcher
+import com.amplitude.android.crash.CrashTrackingRemoteConfig
 import com.amplitude.android.crash.recordCrash
 import com.amplitude.android.diagnostics.AndroidDiagnosticsContextProvider
 import com.amplitude.android.migration.MigrationManager
@@ -61,6 +62,7 @@ public open class Amplitude internal constructor(
                 context = (configuration as Configuration).context,
                 ioDispatcher = storageIODispatcher,
                 diagnosticsClientLazy = lazy { diagnosticsClient },
+                crashTrackingRemoteConfigLazy = lazy { crashTrackingRemoteConfig },
             )
     }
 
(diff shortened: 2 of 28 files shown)


android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -49,6 +51,14 @@ public open class Amplitude internal constructor(
     ) {
     public constructor(configuration: Configuration) : this(configuration, State())
 
+    private val crashCatcher = CrashCatcher(configuration.context)
+
+    init {
+        crashCatcher.consumePreviousCrash()?.let { previousCrash ->
+            diagnosticsClient.recordCrash(previousCrash)
+        }
```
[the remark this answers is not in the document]
ianmeyer: won’t fix — instance replacement is rare; handler chain + consume-on-init is fine

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -49,6 +51,18 @@ public open class Amplitude internal constructor(
     ) {
     public constructor(configuration: Configuration) : this(configuration, State())
 
+    // Assigned in [initWatchers], after configuration validation and before any other
+    // initialization, so the handler covers SDK init without leaking on invalid config.
+    private lateinit var crashCatcher: CrashCatcher
+
+    override fun initWatchers() {
```
sojingle, line 59: iOS gates registration three ways. Sample rate> 0; remote config key `enabled`; and additionally remote-config key `diagnostics.availabilities.CrashTracking` with version-compared. So the crash catcher can be switched off remotely if it ever conflicts with another SDK. It looks like Android is unconditionally enabled here.
Will there also be concerns on Android about compatibility with other SDKs? We currently position crash catching as part of diagnostics, so at minimum it should be constrained by the diagnostics sample rate and enabled setting.

sojingle: For now it will also be limited by sampling at reporting, but we are being conservative during registration mainly to minimize the potential impact on the host app, because diagnostics is the only consumer of crash tracking now.

ianmeyer: We could add a remote config control if that's a feature requirement. I don't know if that's a high-level requirement or just needed on iOS due to SDK conflicts. I don't understand why iOS needed to have a version check so I can't comment on that part.

Our SDK won't conflict with others, because we wrap an existing `UncaughtExceptionHandler`. Will log the exception and then forward to a pre-existing handler. This means that any SDK that initializes before us will not be affected.  We obviously can't guarantee it, but we would expect other SDKs to use the same chaining logic to play nice with others if they initialize after us.

Right now, the exceptions would be forwarded to the diagnostics client, but they would no-op based on the `enabled` and sample rate checks within diagnostics. If I correctly understood the iOS behavior, then the Android gating is the same outcome with a different implementation.

I was intentionally trying to lean away from conditionally initializing the crash catching – delaying the registration of our UncaughtExceptionHandler means that we would miss crashes before registration, which seems dangerous –  thats why `initWatchers()` is introduced – there was no other way to initialize the crash tracking before the `core/Amplitude.init` and `android/Amplitude.build()` had run.

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
```diff
@@ -0,0 +1,56 @@
...
+        Thread.setDefaultUncaughtExceptionHandler { thread, throwable ->
+            try {
+                saveCrashReport(throwable)
+            } catch (_: Throwable) {
+                // Best-effort: never throw from the uncaught exception handler
+            } finally {
+                defaultCrashHandler?.uncaughtException(thread, throwable)
+                    ?: run {
+                        // If no handler to chain to, kill the process
+                        Process.killProcess(Process.myPid())
```
sojingle: Is it better to rethrow and let `ThreadGroup.uncaughtException` take the default path instead of killing the process directly? Killing the process will bypass OS crash reporting and make it like an Amplitude exit rather than a crash.

ianmeyer: throwing here won't work, it would be ignored. This is intentional because then the framework would have to deal with crashes inside our uncaught exception code.  We have our own try/catch so that even if we crash, we still forward the crash to the chain.

Also, in real environments, I think we don't really expect that the `getUncaughtExceptionHandler()` would actually return null.

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Thread.UncaughtExceptionHandler.html#:~:text=Any%20exception%20thrown%20by%20this%20method%20will%20be%20ignored%20by%20the%20Java%20Virtual%20Machine.

android/src/main/java/com/amplitude/android/Amplitude.kt
sojingle: The gates aren't only about SDK conflicts; they follow from how crash tracking is positioned. Right now it's a part of diagnostics, so its value is bounded by diagnostics: for a user with diagnostics disabled, or sampled out at sampleRate = 0 (the default until remote config says otherwise), a captured crash can never be uploaded. Crash tracking has exactly zero value for that population, yet they'd still carry the full cost — a process-wide handler at the front of the crash chain, doing synchronous file I/O at crash time. Gating at reporting vs. gating at registration yields the same data outcome, but the risk we ask every host app to take on is very different. And since sampled-in sessions are the only place a report can ever come from, restricting capture to `enabled && sampleRate > 0` loses nothing.

You're right that this trades away coverage: iOS only registers once remote config arrives, so init-window crashes are missed. That was deliberate — being conservative with a process-wide handler in the first release, since diagnostics is its only consumer today.

On `availabilities.CrashTracking`: it's a version-gated rollout and kill switch. The server enables crash tracking only for SDK versions >= `availableFrom`, so we can ramp gradually — and if a version misbehaves in the field, raising `availableFrom` turns it off for that version with no client release.

That mechanism matters more on Android than it did on iOS. On iOS, crash tracking shipped while diagnostics itself was still ramping, so gating on `sampleRate > 0` gave us a staged rollout for free — crash tracking reached users at exactly the pace diagnostics did. Android diagnostics is already fully rolled out, so the same gate would light up crash tracking for the entire Android population at once. If we want to ramp a process-wide exception handler rather than ship it to everyone on day one, `availabilities` is the only knob left. The kill-switch side is a separate bonus, and how much you value it depends on the confidence in the implementation.

ianmeyer: Sure, I will add gating (including RemoteConfig) before persisting to file in our `UncaughtExceptionHandler`.

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -122,6 +155,12 @@ public open class Amplitude internal constructor(
         // the replacement owns both by then.
         if (!isActive) return
 
+        crashCatcher.consumePreviousCrash()?.let { previousCrash ->
+            AmplitudeRegistry.runIfActive(this) {
+                diagnosticsClient.recordCrash(previousCrash)
+            }
+        }
```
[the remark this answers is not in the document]
ianmeyer: this is fine, we don't expect lots of instances to be created and destroyed and race

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
```diff
@@ -0,0 +1,102 @@
...
+                    // Best-effort: never throw from the uncaught exception handler
+                } finally {
+                    previousHandler?.uncaughtException(thread, throwable)
+                        ?: run {
+                            // If no handler to chain to, kill the process
+                            Process.killProcess(Process.myPid())
+                            exitProcess(10)
+                        }
+                }
+            }
```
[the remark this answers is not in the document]
ianmeyer: Addressed for the paths we care about. `detach()` drops the `Lazy`s the handler captured, and `detachWatchers()` runs from `retire()`, failed `activate()`, and `build()` if construction throws before `activate`. After that the wrapper can stay in the UEH chain without pinning the instance.

We don't WeakRef those lazies — nothing else would own them, so they could be collected while Amplitude is still alive.

An instance that is constructed and dropped without retire/replace would still be reachable through the handler; that isn't expected SDK use.

android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -49,6 +52,20 @@ public open class Amplitude internal constructor(
...
+    private lateinit var crashCatcher: CrashCatcher
+
+    override fun initWatchers() {
+        crashCatcher =
+            CrashCatcher(
+                context = (configuration as Configuration).context,
+                ioDispatcher = storageIODispatcher,
+                diagnosticsClientLazy = lazy { diagnosticsClient },
+                crashTrackingRemoteConfigLazy = lazy { crashTrackingRemoteConfig },
+            )
```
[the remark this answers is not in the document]
ianmeyer: Same leak as the later thread on `CrashCatcher.detach()`. Addressed: `detach()` drops the captured `Lazy`s, and `detachWatchers()` runs from `retire()`, failed `activate()`, and `build()` if construction throws before `activate`. WeakRef on those lazies would let them be collected while Amplitude is still alive.

analytics-core/src/main/java/com/amplitude/core/diagnostics/DiagnosticsClientImpl.kt
```diff
@@ -46,8 +46,9 @@ internal class DiagnosticsClientImpl(
     private val startTimestampSeed: String = startTimestamp.toString()
     private var sampleRate: Double = sampleRate.coerceIn(0.0, 1.0)
 
-    private var shouldTrack: Boolean =
+    override var shouldTrack: Boolean =
         enabled && Sample.isInSample(seed = startTimestampSeed, sampleRate = this.sampleRate)
+        private set
```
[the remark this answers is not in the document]
ianmeyer: Fixed. `shouldTrack` is `@Volatile` so the uncaught-exception handler sees updates from the diagnostics actor.

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
```diff
@@ -0,0 +1,81 @@
...
+    init {
+        Thread.setDefaultUncaughtExceptionHandler { thread, throwable ->
+            try {
+                if (
+                    throwable.isAmplitudeSdkCrash() &&
+                    crashTrackingRemoteConfigLazy.value.isCrashTrackingEnabled &&
+                    diagnosticsClientLazy.value.shouldTrack
+                ) {
+                    saveCrashReport(throwable)
+                }
```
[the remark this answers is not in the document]
ianmeyer: Leaving the SDK-frame check first for now. When tracking is off we still walk the stack, then fail the remote-config / `shouldTrack` gates. Acceptable for v1; we can cheap-gate first in a follow-up if the crash-path cost matters.

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
```diff
@@ -0,0 +1,82 @@
...
```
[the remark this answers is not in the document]
ianmeyer: Intentional for v1. Persist is gated on the crashing session's `shouldTrack` so we do not write crash files for unsampled sessions (default sample rate is 0 until remote config). That can miss crashes before sampling resolves; we accepted that to keep the first ship conservative. After consume, `recordCrash` still buffers for the new session.

android/src/main/java/com/amplitude/android/Amplitude.kt
ianmeyer: Added gating

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
```diff
@@ -0,0 +1,102 @@
...
+            lastPersistedThrowable = throwable
+            crashStorage.saveCrashReport(throwable)
+        }
+    }
+
+    private fun Throwable.isAmplitudeSdkCrash(): Boolean {
+        var current: Throwable? = this
+        val seen = mutableSetOf<Throwable>()
+        while (current != null && seen.add(current)) {
+            if (current.stackTrace.any { it.className.startsWith(AMPLITUDE_PACKAGE_PREFIX) }) {
```
sojingle, line 89: `consumer-rules.pro` doesn't keep all of the Amplitude SDK class names generally, so it will miss crashes caused by classes not included in the ProGuard rules.

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
```diff
@@ -0,0 +1,102 @@
...
+            val previousHandler = Thread.getDefaultUncaughtExceptionHandler()
+            Thread.setDefaultUncaughtExceptionHandler { thread, throwable ->
+                try {
+                    if (
+                        throwable.isAmplitudeSdkCrash() &&
+                        crashTrackingRemoteConfigProvider?.value?.isCrashTrackingEnabled == true &&
+                        // Persist only when this session is already sampled in. That misses
+                        // crashes before remote config arrives (default sample rate is 0), but
+                        // avoids writing crash files in unsampled sessions.
+                        diagnosticsClientProvider?.value?.shouldTrack == true
```
sojingle, line 43: shouldTrack is "is this session sampled in", but a crash is written in session N and can only be uploaded in session N+1 — and upload is already gated on N+1's shouldTrack. Gating the write on N's shouldTrack too makes the effective coverage sampleRate²: at a 0.1 sample rate that's ~1%, not ~10%.

This is the case the iOS implementation has commented about:

> // We intentionally use `sampleRate > 0` instead of `shouldTrack` here.
> // Crashes are captured during one session but can only be reported in the next session.
> // Using `shouldTrack` would require both consecutive sessions to be sampled in,
> // significantly reducing crash coverage (e.g., 1% with a 0.1 sample rate).
> // By checking `sampleRate > 0`, we capture crashes whenever the feature is enabled,
> // and reporting depends only on the next session being sampled in (~10% with 0.1 rate).

The predicate we want is enabled && sampleRate > 0, which still gives us everything we discussed —nothing is written for users who have diagnostics off or a zero sample rate — without the squared decay. `sampleRate` is a private var in DiagnosticsClientImpl today, so this probably means exposing something like `isSamplingEnabled` rather than `shouldTrack`.

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
```diff
@@ -0,0 +1,102 @@
...
+    private var diagnosticsClientProvider: Lazy<DiagnosticsClient>? = diagnosticsClientLazy
+    private var crashTrackingRemoteConfigProvider: Lazy<CrashTrackingRemoteConfig>? = crashTrackingRemoteConfigLazy
+
+    init {
+        synchronized(handlerInstallLock) {
+            val previousHandler = Thread.getDefaultUncaughtExceptionHandler()
+            Thread.setDefaultUncaughtExceptionHandler { thread, throwable ->
+                try {
+                    if (
+                        throwable.isAmplitudeSdkCrash() &&
```
sojingle, line 38: `isAmplitudeSdkCrash` seems like a relatively heavy operation. Putting it in a handler might affect whether the file can be saved in time. It may be more reasonable to perform the check before upload instead.

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
ianmeyer: Okay, I see. I'm not used to sampling being determined by the client, rather than cloud.

Yes, the visibility is a constant friction.

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
ianmeyer: Hmm. We could add `-keeppackagenames` but I guess we're not going to be able to retrace because we have the consuming app obfuscate our code. I'm just going to drop this filter entirely for now and discuss the long-term goals outside this PR.

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
ianmeyer: Is there a specific reason you don't want to expose `sampleRate`? None of it should be exposed but we're stuck with it.

sojingle: no issue with exposing it; both ok for me

android/src/main/java/com/amplitude/android/crash/CrashCatcher.kt
sojingle: If it’s not that reliable, maybe we can attach it as a reference attribute instead of a local filter.

PR #472, opened 2026-08-20
feat: capture Android ANRs for diagnostics

ianmeyer:
Stacked on #471. Capture Android ANRs for SDK diagnostics so a prior-session hang can be reported with the same schema as crash (`analytics.anr` counter + event, `report` property).

### Describe the solution
Two catchers behind `createAnrCatcher()`:
* **API 30+:** `ApplicationExitInfo` unread `REASON_ANR` exits, claimed once under a process-wide mutex so concurrent Amplitude instances do not double-count.
* **Pre-30:** main-thread watchdog that pings every 2s and persists if a ping is unanswered for 5s.

Watchers are installed in `initWatchers()` next to crash capture. Consume is suspend I/O on the storage dispatcher at the call site; persist on the watchdog stays synchronous. A pending legacy report after an OS upgrade to 11+ is not drained (one-off).

* `./gradlew :android:testDebugUnitTest --tests com.amplitude.android.anr.*`
* Sample app **ANR** button (`Thread.sleep(10_000)` on main) on an API 29 emulator, then relaunch and confirm `analytics.anr` in diagnostics when sampled in.
* On API 31+, force an ANR, relaunch, confirm AEI-backed report.

Depends on #471.

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Introduces a process-wide background watchdog and startup I/O on all API levels; reports can contain stack traces, similar to existing crash diagnostics but with extra runtime overhead on pre-Android 11.
> 
> **Overview**
> Adds **prior-session ANR reporting** to the Android SDK, wired like existing crash capture: an `AnrCatcher` is started in `initWatchers()`, reports are consumed during `buildInternal()`, and each is sent via `diagnosticsClient.recordAnr` as an **`analytics.anr`** counter plus event with a **`report`** property.
> 
> **API 30+** uses `ApplicationExitInfo` (`REASON_ANR`), tracks the last consumed exit timestamp so each ANR is reported once (mutex for concurrent Amplitude instances), and includes bounded trace text. **Below API 30**, a process-wide main-thread watchdog pings every 2s and synchronously persists a truncated thread dump after ~5s of no response; the legacy file is read on the next launch.
> 
> `detachWatchers()` now tears down the legacy watchdog when the owning instance retires. The sample app gains an **ANR** button that blocks the main thread for manual testing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9b91ae295cf29e56e9cbe8f4293c8e91ea59fc8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -2,6 +2,9 @@ package com.amplitude.android
 
 import android.app.Application
 import android.content.Context
+import com.amplitude.android.anr.AnrCatcher
+import com.amplitude.android.anr.createAnrCatcher
+import com.amplitude.android.anr.recordAnr
 import com.amplitude.android.crash.CrashCatcher
 import com.amplitude.android.crash.CrashTrackingRemoteConfig
 import com.amplitude.android.crash.recordCrash
@@ -55,11 +58,18 @@ public open class Amplitude internal constructor(
     // Assigned in [initWatchers], after configuration validation and before any other
     // initialization, so the handler covers SDK init without leaking on invalid config.
     private lateinit var crashCatcher: CrashCatcher
+    private lateinit var anrCatcher: AnrCatcher
 
     override fun initWatchers() {
+        val androidConfig = configuration as Configuration
         crashCatcher =
             CrashCatcher(
-                context = (configuration as Configuration).context,
+                context = androidConfig.context,
+                ioDispatcher = storageIODispatcher,
+            )
+        anrCatcher =
+            createAnrCatcher(
+                context = androidConfig.context,
                 ioDispatcher = storageIODispatcher,
                 diagnosticsClientLazy = lazy { diagnosticsClient },
                 crashTrackingRemoteConfigLazy = lazy { crashTrackingRemoteConfig },
@@ -159,6 +169,9 @@ public open class Amplitude internal constructor(
             AmplitudeRegistry.runIfActive(this) {
                 diagnosticsClient.recordCrash(previousCrash)
             }
+        }
+        anrCatcher.consumePreviousAnrs().forEach { previousAnr ->
+            diagnosticsClient.recordAnr(previousAnr)
         }
 
         val migrationManager = MigrationManager(this)
--- a/android/src/main/java/com/amplitude/android/Amplitude.kt
+++ b/android/src/main/java/com/amplitude/android/Amplitude.kt
@@ -255,6 +255,10 @@ public open class Amplitude internal constructor(
             runCatching { crashCatcher.detach() }
                 .onFailure { logger.warn("Failed to detach the crash handler: $it") }
         }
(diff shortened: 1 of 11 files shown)


android/src/main/java/com/amplitude/android/Amplitude.kt
```diff
@@ -55,15 +58,22 @@ public open class Amplitude internal constructor(
...
         crashCatcher =
             CrashCatcher(
-                context = (configuration as Configuration).context,
+                context = androidConfig.context,
                 ioDispatcher = storageIODispatcher,
                 diagnosticsClientLazy = lazy { diagnosticsClient },
                 crashTrackingRemoteConfigLazy = lazy { crashTrackingRemoteConfig },
             )
+        anrCatcher =
+            createAnrCatcher(
```
sojingle, line 73: Compared with CrashCatcher, ANR seems to have even higher runtime overhead, so it should probably also be gated like CrashCatcher (enabled, sampleRate > 0, availabilities)

android/src/main/java/com/amplitude/android/anr/LegacyAnrCatcher.kt
```diff
@@ -0,0 +1,140 @@
...
+                    val running = watchdogThread
+                    watchdogThread = null
+                    activeCatcher = null
+                    running
+                }
+            thread?.interrupt()
+            thread?.join(1_000)
+        }
+
+        private fun runWatchdog() {
```
sojingle, line 97: Should we consider stopping continuous monitoring after switching to the background?

android/src/main/java/com/amplitude/android/anr/AnrStorage.kt
```diff
@@ -0,0 +1,136 @@
...
+                } catch (_: Exception) {
+                    0L
+                }
+            }
+        }
+
+    suspend fun saveLastAeiTimestamp(timestamp: Long) {
+        withContext(ioDispatcher) {
+            synchronized(fileLock) {
+                File(directory, AEI_TIMESTAMP_FILE_NAME).writeText(timestamp.toString())
```
sojingle, line 86: try/catch this?

android/src/main/java/com/amplitude/android/anr/AnrStorage.kt
```diff
@@ -0,0 +1,136 @@
...
+internal class AnrStorage(
+    private val appContext: Context,
+    private val ioDispatcher: CoroutineDispatcher,
+) {
+    val directory: File by lazy {
+        File(
+            appContext.getDir("amplitude", Context.MODE_PRIVATE),
+            STORAGE_PREFIX,
+        ).also {
+            it.mkdirs()
```
sojingle, line 28: Should try/catch this. If it throws, it has a chance to stop the SDK from initializing through `consumePreviousAnr`

PR #483, opened 2026-09-02
feat: Streaming - Module

ianmeyer:
### Describe what this PR is addressing
[SDKA-85](https://linear.app/amplitude/issue/SDKA-85/create-streaming-analytics-android-module): add an empty `:streaming-analytics-android` library so later PRs can land tracking, delayed events, and publishing on top of a stable Gradle module.

Stacked on [#482](https://github.com/amplitude/Amplitude-Kotlin/pull/482).

### Describe the solution
Adds `:streaming-analytics-android` (`com.amplitude:streaming-analytics-android`) as an unpublished Android library:
* Gradle project, version catalog deps, `explicitApi()`, and a commented Maven publish config to turn on when the public API is ready
* Empty `consumer-rules.pro` placeholder (no blanket `-keepnames`)
* Unused `AmplitudePreview` opt-in in `analytics-core` for later preview APIs

No tracking implementation yet. The sample app is unchanged.

### Steps to verify the change
1. `./gradlew :streaming-analytics-android:assembleDebug`
2. `./gradlew ktlintCheck apiCheck`

### Useful links and documentation (optional)
* [SDKA-85](https://linear.app/amplitude/issue/SDKA-85/create-streaming-analytics-android-module)
* [#482](https://github.com/amplitude/Amplitude-Kotlin/pull/482)

### Checklist
* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-Kotlin/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* [x] Does your PR have a breaking change?

---

> [!NOTE]
> **Low Risk**
> Scaffolding and additive API (preview annotation) only; no event pipeline or publishing behavior changes yet.
> 
> **Overview**
> Introduces an unpublished **`:streaming-analytics-android`** Android library (Media3 / ExoPlayer–oriented deps, `explicitApi()`, commented Maven publish) so follow-up PRs can land streaming tracking on a fixed Gradle module. **`settings.gradle.kts`** wires the new project; the version catalog and root **`build.gradle.kts`** add Kotlin serialization plus Media3/IMA library aliases.
> 
> **`analytics-core`** gains a new public **`AmplitudePreview`** `@RequiresOptIn` annotation (reflected in the API dump) for marking unstable streaming APIs later. **`.gitignore`** now ignores **`.kotlin/`** and **`.cursor/`**.
> 
[...]

PR #485, opened 2026-09-02
feat: Streaming - Public API

ianmeyer:
[SDKA-90](https://linear.app/amplitude/issue/SDKA-90/public-api-trackplayer-classpath-autocapture): preview `Amplitude.trackPlayer` and classpath install of Streaming Analytics when the artifact is present.

Stacked on [#483](https://github.com/amplitude/Amplitude-Kotlin/pull/483).

### Describe the solution
Adds the preview public surface on `:streaming-analytics-android`:
* `Amplitude.trackPlayer(player, contentProvider)` (`@AmplitudePreview`). Java: `AmplitudeStreamingAnalytics.trackPlayer(...)`
* `PlayerContent` / `PlayerContentProvider`
* `StreamingAnalyticsPlugin`, installed from `:android` via `Class.forName` during `buildInternal` when this module is on the classpath
* Internal `StreamingAnalytics` per Amplitude instance. `trackPlayer` / `teardown` are stubs except that `teardown` unregisters the instance so Amplitude replacements do not leak

Player observation, delayed events, and heartbeat stay on a follow-up commit.

1. `./gradlew :streaming-analytics-android:test :android:testDebugUnitTest --tests com.amplitude.android.plugins.OptionalClasspathPluginsTest`
2. `./gradlew ktlintCheck apiCheck`
3. Sample `:samples:streaming-app` compiles against `trackPlayer` (no-op tracking)

* [SDKA-90](https://linear.app/amplitude/issue/SDKA-90/public-api-trackplayer-classpath-autocapture)
* [#483](https://github.com/amplitude/Amplitude-Kotlin/pull/483)

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Reflective classpath plugin installation and new Enrichment pipeline behavior affect SDK startup and event routing; tracking logic is stubbed so functional risk is lower until follow-up work lands.
> 
> **Overview**
> Introduces the **preview** (`@AmplitudePreview`) public surface on `:streaming-analytics-android` for Media3 streaming analytics: `Amplitude.trackPlayer` (Java: `AmplitudeStreamingAnalytics.trackPlayer`), `PlayerContent`, `PlayerContentProvider`, and `StreamingAnalyticsPlugin`. Player observation and event emission are **not implemented yet**—internal `StreamingAnalytics` methods are stubs, and the plugin only wires lifecycle plus **Enrichment** handling that consumes internal `DelayedEvent`s before destinations.
> 
> When the streaming artifact is on the classpath, `:android` now **reflectively installs** `StreamingAnalyticsPlugin` during `buildInternal` via new `OptionalClasspathPlugins` (no-op if absent; errors logged otherwise). R8 consumer rules keep the plugin’s no-arg constructor for that path. `trackPlayer` can also register the plugin on demand, with logic to respect Amplitude’s first-name-wins `add` behavior.
> 
> Build/docs: binary-compat validation is extended for this module (manual `apiDump`/`apiCheck` on AGP 9, BCV dependency in `buildSrc`, committed `streaming-analytics-android.api`), and the streaming sample calls `trackPlayer` after creating demo players.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4061b565cb791b76715c39b885562486bca436f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
+++ b/streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
@@ -5,7 +5,6 @@ package com.amplitude.android
 import androidx.media3.common.Player
 import com.amplitude.android.streaming.PlayerContentProvider
 import com.amplitude.android.streaming.StreamingAnalyticsPlugin
-import com.amplitude.core.Amplitude
 import com.amplitude.core.AmplitudePreview
 
 /**
--- a/streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
+++ b/streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
@@ -1,0 +1,39 @@
+@file:JvmName("AmplitudeStreamingAnalytics")
+
+package com.amplitude.android
+
+import androidx.media3.common.Player
+import com.amplitude.android.streaming.PlayerContentProvider
+import com.amplitude.android.streaming.StreamingAnalyticsPlugin
+import com.amplitude.core.Amplitude
+import com.amplitude.core.AmplitudePreview
+
+/**
+ * Starts Streaming Analytics on [androidx.media3.common.Player].
+ *
+ * If [player] is already tracked by this Amplitude instance, this function is a no-op
+ * and [contentProvider] is ignored.
+ *
+ * ```
+ * amplitude.trackPlayer(exoPlayer) { mediaItem ->
+ *     PlayerContent(
+ *         contentId = mediaItem?.mediaId ?: "ep-1",
+ *         title = "Episode 1",
+ *     )
+ * }
+ * ```
+ */
+@AmplitudePreview
+public fun Amplitude.trackPlayer(
+    player: Player,
+    contentProvider: PlayerContentProvider,
+) {
+    val plugin =
+        findPlugin<StreamingAnalyticsPlugin>()
+            ?: StreamingAnalyticsPlugin().also { add(it) }
+
+    plugin.streamingAnalytics?.trackPlayer(
+        player = player,
+        contentProvider = contentProvider,
+    )
+}
--- a/streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
+++ b/streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
@@ -28,12 +28,26 @@ public fun Amplitude.trackPlayer(
     player: Player,
     contentProvider: PlayerContentProvider,
 ) {
-    val plugin =
-        findPlugin<StreamingAnalyticsPlugin>()
-            ?: StreamingAnalyticsPlugin().also { add(it) }
-
+    val plugin = resolveStreamingAnalyticsPlugin(findPlugin())
     plugin.streamingAnalytics?.trackPlayer(
         player = player,
         contentProvider = contentProvider,
     )
+}
+
+/**
+ * Returns the registered [StreamingAnalyticsPlugin], installing one if needed.
+ *
+ * Plugin registration is first-name-wins. If another thread already registered the
+ * plugin, [Amplitude.add] keeps that instance; resolve it instead of using a rejected
+ * candidate that was never set up.
+ */
+@OptIn(AmplitudePreview::class)
+internal fun Amplitude.resolveStreamingAnalyticsPlugin(
+    existing: StreamingAnalyticsPlugin?,
+): StreamingAnalyticsPlugin {
+    existing?.let { return it }
+    val candidate = StreamingAnalyticsPlugin()
+    add(candidate)
+    return plugin(candidate.name) as? StreamingAnalyticsPlugin ?: candidate
 }
--- a/streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
+++ b/streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
@@ -6,6 +6,7 @@ import androidx.media3.common.Player
 import com.amplitude.android.streaming.PlayerContentProvider
 import com.amplitude.android.streaming.StreamingAnalyticsPlugin
 import com.amplitude.core.AmplitudePreview
+import kotlinx.coroutines.launch
 
 /**
  * Starts Streaming Analytics on [androidx.media3.common.Player].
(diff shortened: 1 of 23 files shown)


streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
```diff
@@ -0,0 +1,53 @@
...
+ * amplitude.trackPlayer(exoPlayer) { mediaItem ->
+ *     PlayerContent(
+ *         contentId = mediaItem?.mediaId ?: "ep-1",
+ *         title = "Episode 1",
+ *     )
+ * }
+ * ```
+ */
+@AmplitudePreview
+public fun Amplitude.trackPlayer(
```
polbins, line 27: Can this target `com.amplitude.android.Amplitude` instead of core Amplitude? Since it's an Android feature, just so it doesn't get mixed up.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingAnalytics.kt
```diff
@@ -0,0 +1,30 @@
...
+@Suppress("UNUSED_PARAMETER")
+internal class StreamingAnalytics(
+    private val amplitude: Amplitude,
+) {
+    fun trackPlayer(
+        player: Player,
+        contentProvider: PlayerContentProvider,
+    ) {
+        // TODO: Not yet implemented
+    }
```
polbins, line 21: Should we provide `untrackPlayer(player)`? I'm assuming consumers would want a way to detach to a player individually.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
polbins: Oh, it seems to be fixed here: https://github.com/amplitude/Amplitude-Kotlin/pull/486/changes#diff-cfd7ffb4780101d2535858f45624a8b33a6c259cba4c6e017c40d9463b6aca52R26

streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Returns the registered [StreamingAnalyticsPlugin], installing one if needed.
+ *
+ * Plugin registration is first-name-wins. If another thread already registered the
+ * plugin, [Amplitude.add] keeps that instance; resolve it instead of using a rejected
+ * candidate that was never set up.
+ */
+@OptIn(AmplitudePreview::class)
+internal fun Amplitude.resolveStreamingAnalyticsPlugin(
```
polbins:
```suggestion
@OptIn(AmplitudePreview::class)
@JvmSynthetic
internal fun Amplitude.resolveStreamingAnalyticsPlugin(
```

polbins (commented):
Can we get an `apiDump` generated for this module please? It could helpe make it easier to see which ones are added/modified starting from this PR.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingAnalytics.kt
ianmeyer: Yeah, I'd like to sync with the other platforms and not aggressively add it or block an MVP on this. The options are:
1. Don't have an untrack in the API. It's not totally clear to me that it's required.
2. Return an handle object from this method that exposes a stop API.
3. Create `Amplitude.untrackPlayer(...)`

streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
ianmeyer: Why would we need `@JvmSynthetic` here? We don't want anyone to call this directly from Java.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingAnalytics.kt
polbins: Thanks for explaining! Option 1 works for the MVP, and then maybe we can revisit Option 2/3 once we have other platforms settle on the lifecycle

streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
polbins: That’s the reason for it, as Kotlin internal is still emitted as a public JVM method, so Java callers can invoke it. `@JvmSynthetic` hides it from Java source

streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
```diff
@@ -0,0 +1,53 @@
...
+ *     )
+ * }
+ * ```
+ */
+@AmplitudePreview
+public fun Amplitude.trackPlayer(
+    player: Player,
+    contentProvider: PlayerContentProvider,
+) {
+    val plugin = resolveStreamingAnalyticsPlugin(findPlugin())
```
polbins: I think the automatic installation introduces a race here. 

Thread A: 

> plugin reserved -> setup() running -> ready

Thread B:        

> findPlugin() = null -> add() loses as plugin isn't ready
>                                         -> trackPlayer() is dropped

I was able to reproduce this on a unit test by pausing `setup()`. 

Can we keep the usual plugin lifecycle and require `amplitude.add(StreamingAnalyticsPlugin())` once before `trackPlayer()`?

streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
ianmeyer: In any case, this method is removed now. 

That said, we aren't requiring that we add `@JvmSynthetic` to internals elsewhere, so maybe that would be a larger discussion outside of this PR if we wanted to do that.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
ianmeyer: Yeah, this was a bit over-engineered in the wrong direction. It now is async and awaits isBuilt so the plugin will be installed.

polbins: The `isBuilt.await()` part works. Thanks!

Now I've tried the new `trackPlayer(...)` on the streaming-app and it crashed with:
image

It looks related to the change to `amplitudeDispatcher`, it seems the Player should only be accessed from the main thread.

polbins: this is fixed in #492

polbins (approved):

PR #486, opened 2026-09-02
feat: Streaming - Network

ianmeyer:
[SDKA-86](https://linear.app/amplitude/issue/SDKA-86/delayed-events-http-client): delayed-events HTTP client (`POST /2/httpapi/delayed`).

Stacked on [#485](https://github.com/amplitude/Amplitude-Kotlin/pull/485).

### Describe the solution
Adds an internal HTTP client in `:streaming-analytics-android` for delayed events:
* Host from `serverUrl` when set, otherwise US/EU from `serverZone` (same pattern as session-replay-android)
* `DelayedEventsEndpoint.send` POSTs JSON; 2xx is success, `429` is non-fatal rate-limit, other statuses and connection errors are failure
* DTOs + Metro `DiGraph` wiring for the endpoint

Player observation and heartbeat stay on a follow-up commit.

1. `./gradlew :streaming-analytics-android:test --tests com.amplitude.android.streaming.internal.network.AmplitudeBaseUrlTest --tests com.amplitude.android.streaming.internal.network.DelayedEventsEndpointTest`
2. `./gradlew ktlintCheck apiCheck`

* [SDKA-86](https://linear.app/amplitude/issue/SDKA-86/delayed-events-http-client)
* [#485](https://github.com/amplitude/Amplitude-Kotlin/pull/485)

### Checklist

---

> [!NOTE]
> **Medium Risk**
> New outbound analytics HTTP path and event JSON serialization affect data sent off-device; wiring into playback is still pending, but misconfigured URLs or payloads could drop or mis-route delayed events.
> 
> **Overview**
> Adds an **internal delayed-events HTTP stack** in `:streaming-analytics-android`: **`DelayedEventsBaseUrl`** picks `serverUrl` or US/EU delayed-events hosts and appends path segments; **`DelayedEventsEndpoint.send`** POSTs JSON to **`/delayed`**, mapping **2xx** to success, **429** to rate-limited, and other/transport errors to **`DelayedEventsResult.Failure`**. Request bodies use **`DelayedEventsRequestDto`** (`id`, `timeout`, `events`, optional **`instant_events`**) with events serialized through an updated **`DelayedEvent`** (**`Kind`**, wrap-from-**`BaseEvent`** with deep-copied maps, **`DelayedEventSerializer`** via **`JSONUtil`**).
> 
> Introduces a reusable **`DiGraph`** (singleton/weak lazy deps) and **`StreamingDiGraph`** to expose configuration, IO scope, and weak **`delayedEventsBaseUrl`** / **`delayedEventsEndpoint`**. **`StreamingAnalyticsPlugin.setup`** now **requires `com.amplitude.android.Amplitude`**; tests use mocked Android **`Amplitude`**. Player queuing/heartbeat still stubbed in **`StreamingAnalytics`**—this PR is the network + serialization foundation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d1c0da992a7043eeb981291058f5fad72816d3d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingDiGraph.kt
+++ b/streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingDiGraph.kt
@@ -1,0 +1,33 @@
+package com.amplitude.android.streaming.internal
+
+import android.content.Context
+import com.amplitude.android.Configuration
+import com.amplitude.core.Amplitude
+import com.amplitude.android.streaming.internal.util.DiGraph
+import com.amplitude.common.Logger
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+
+/**
+ * Dependency graph for streaming analytics.
+ *
+ * Dependencies are declared as extension properties on this class in the files that own
+ * their types.
+ */
+internal class StreamingDiGraph(
+    val amplitude: Amplitude,
+) : DiGraph() {
+
+    val configuration: Configuration by lazy { amplitude.configuration as Configuration }
+    val context: Context by lazy { configuration.context.applicationContext }
+    val ioDispatcher: CoroutineDispatcher by lazy { Dispatchers.IO }
+    val logger: Logger by lazy { amplitude.logger }
+    val scope: CoroutineScope by lazy {
+        CoroutineScope(
+            SupervisorJob(amplitude.amplitudeScope.coroutineContext[Job]) + ioDispatcher,
+        )
+    }
+}
--- a/streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsBaseUrl.kt
+++ b/streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsBaseUrl.kt
@@ -25,13 +25,28 @@ internal class DelayedEventsBaseUrl(
                     ServerZone.EU -> EU_DEFAULT_HOST
                 }
         val uri = URI(base)
-        val basePath = uri.rawPath.orEmpty().trimEnd('/').ifEmpty { "/2/httpapi" }
+        val basePath = uri.rawPath.orEmpty().trimEnd('/')
         val extra =
             extraPathSegments
                 .map { it.trim('/') }
                 .filter { it.isNotEmpty() }
                 .joinToString("/")
         val path = if (extra.isEmpty()) basePath else "$basePath/$extra"
-        return URL(URI(uri.scheme, uri.authority, path, uri.query, uri.fragment).toString())
+        return URL(
+            buildString {
+                append(uri.scheme)
+                append("://")
+                append(uri.rawAuthority)
+                append(path)
+                if (!uri.rawQuery.isNullOrEmpty()) {
+                    append('?')
+                    append(uri.rawQuery)
+                }
+                if (!uri.rawFragment.isNullOrEmpty()) {
+                    append('#')
+                    append(uri.rawFragment)
+                }
+            },
+        )
     }
 }
--- a/streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsBaseUrl.kt
+++ b/streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsBaseUrl.kt
@@ -3,24 +3,26 @@ package com.amplitude.android.streaming.internal.network
 import com.amplitude.android.streaming.internal.StreamingDiGraph
 import com.amplitude.android.streaming.internal.util.DiGraph.Companion.weak
 import com.amplitude.core.Configuration
-import com.amplitude.core.Constants
 import com.amplitude.core.ServerZone
 import java.net.URI
 import java.net.URL
 
-internal val StreamingDiGraph.amplitudeBaseUrl: AmplitudeBaseUrl by weak {
-    AmplitudeBaseUrl(configuration)
+private const val US_DEFAULT_HOST = "https://delayed-events.prod.us-west-2.amplitude.com/2/httpapi"
+private const val EU_DEFAULT_HOST = "https://delayed-events.prod.eu-central-1.amplitude.com/2/httpapi"
+
+internal val StreamingDiGraph.delayedEventsBaseUrl: DelayedEventsBaseUrl by weak {
+    DelayedEventsBaseUrl(configuration)
 }
 
-internal class AmplitudeBaseUrl(
+internal class DelayedEventsBaseUrl(
     private val configuration: Configuration,
 ) {
     fun url(vararg extraPathSegments: String): URL {
         val base =
             configuration.serverUrl?.takeIf { it.isNotBlank() }
                 ?: when (configuration.serverZone) {
-                    ServerZone.US -> Constants.DEFAULT_API_HOST
-                    ServerZone.EU -> Constants.EU_DEFAULT_API_HOST
+                    ServerZone.US -> US_DEFAULT_HOST
+                    ServerZone.EU -> EU_DEFAULT_HOST
                 }
         val uri = URI(base)
         val basePath = uri.rawPath.orEmpty().trimEnd('/').ifEmpty { "/2/httpapi" }
--- a/streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsEndpoint.kt
+++ b/streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsEndpoint.kt
@@ -6,6 +6,7 @@ import com.amplitude.common.Logger
 import com.amplitude.core.Configuration
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.withContext
+import kotlinx.serialization.SerializationException
 import java.io.IOException
 import java.io.InputStream
 import java.net.HttpURLConnection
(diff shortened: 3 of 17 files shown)


streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingDiGraph.kt
```diff
@@ -0,0 +1,33 @@
...
+ * Dependency graph for streaming analytics.
+ *
+ * Dependencies are declared as extension properties on this class in the files that own
+ * their types.
+ */
+internal class StreamingDiGraph(
+    val amplitude: Amplitude,
+) : DiGraph() {
+
+    val configuration: Configuration by lazy { amplitude.configuration as Configuration }
```
polbins, line 24: Should we restrict this to `com.amplitude.android`? Cast here would fail if we have a `com.amplitude.core` Configuration supplied

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsEndpoint.kt
```diff
@@ -0,0 +1,85 @@
...
+            try {
+                connection.requestMethod = "POST"
+                connection.connectTimeout = CONNECT_TIMEOUT_MILLIS
+                connection.readTimeout = READ_TIMEOUT_MILLIS
+                connection.doInput = true
+                connection.doOutput = true
+                connection.setRequestProperty("Content-Type", "application/json; charset=utf-8")
+                connection.setRequestProperty("Accept", "application/json")
+                connection.outputStream.use {
+                    it.write(request.toJson(configuration.apiKey).toByteArray(Charsets.UTF_8))
```
polbins: Can we remove the `requestBody` outside and catch `SerializationException` as well? As `encodeToString` can throw (and there's also an `OptIn` warning on `DelayedEventsRequestDto`).

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsEndpoint.kt
```diff
@@ -0,0 +1,85 @@
...
+                }
+                val statusCode = connection.responseCode
+                val responseBody = readBody(connection)
```
polbins: Can we read the `statusCode` before reading `responseBody`? It's only used in failure, and if there's something wrong with reading (e.g. read timeout), it can stall and then become an exception/failure.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsBaseUrl.kt
```diff
@@ -0,0 +1,35 @@
...
+) {
+    fun url(vararg extraPathSegments: String): URL {
+        val base =
+            configuration.serverUrl?.takeIf { it.isNotBlank() }
+                ?: when (configuration.serverZone) {
+                    ServerZone.US -> Constants.DEFAULT_API_HOST
+                    ServerZone.EU -> Constants.EU_DEFAULT_API_HOST
+                }
+        val uri = URI(base)
+        val basePath = uri.rawPath.orEmpty().trimEnd('/').ifEmpty { "/2/httpapi" }
```
polbins: I believe the `serverUrl` is used to replace the batch endpoint and is used as-is. Should we just append `/delayed`, and not `/2/httpapi/delayed`?

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingDiGraph.kt
```diff
@@ -0,0 +1,33 @@
...
+    val context: Context by lazy { configuration.context.applicationContext }
+    val ioDispatcher: CoroutineDispatcher by lazy { Dispatchers.IO }
```
polbins, line 26: could we use `amplitude.networkIODispatcher` here?

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingDiGraph.kt
ianmeyer: How could we restrict it?  Even if StreamingAnalytics takes `com.amplitude.android.Amplitude`, there still needs to be a cast here.  The plugin is passing `com.amplitude.core.Amplitude`, so there'd still be a cast to get to  `com.amplitude.android.Amplitude`.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsBaseUrl.kt
ianmeyer: I'm not sure. This is something where I probably need to circle back when I know the final endpoint location.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingDiGraph.kt
ianmeyer: `amplitude.networkIODispatcher` is using a dispatcher to act like a single thread, it's not really using coroutines as intended. I don't think we want to continue to do that for greenfield code.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsEndpoint.kt
ianmeyer: sure. done.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsEndpoint.kt
ianmeyer: sure. done.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingDiGraph.kt
polbins: I agree that `Plugin.setup` has to keep the core type because of the interface. But since this is an Android-only plugin, maybe we should **fail early** just in case it was misconfigured and then use the Android specific classes moving forward.

something like:
```
override fun setup(amplitude: CoreAmplitude) {
    require(amplitude is AndroidAmplitude) {
        "StreamingAnalyticsPlugin requires com.amplitude.android.Amplitude"
    }
    super.setup(amplitude)
    streamingAnalytics = StreamingAnalytics(amplitude)
}
```

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsBaseUrl.kt
polbins: Gotcha 👍🏽 maybe we can keep this open until this is clear.

ianmeyer: updated

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingDiGraph.kt
ianmeyer: Yeah, done. We still have to cast internally unless Plugin is made generic.

polbins: Thanks! I can see the early guard now, but I think we should carry the `com.amplitude.android.Amplitude` Android type forward after it. This should remove the need for casting the `Configuration` and such when we implement more Android specific features down the line.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsBaseUrl.kt
polbins: Could we preserve the raw path and query when rebuilding this URL?  `/a%2Fb?token=x%2Fy` became `/a%252Fb/delayed?token=x/y`

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsBaseUrl.kt
polbins: So the base URLs are updated, but then the usage of custom `serverUrl` is still going to append `/2/httpapi`.

ianmeyer: Okay. I think the requirements are just not really clear: https://amplitude.atlassian.net/wiki/spaces/Middleware/pages/3775988254/Delayed+Events+Endpoint+aka+heartbeat+service#Endpoints

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingDiGraph.kt
ianmeyer: `com.amplitude.android.Android.configuration` is type `com.amplitude.core.Android.configuration`, so a cast cannot be avoided here.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/network/DelayedEventsBaseUrl.kt
ianmeyer: updated, with a test

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/StreamingDiGraph.kt
polbins: Oh no.. But at least the StreamingAnalytics is on Android now 👍🏽

polbins (approved):

PR #502, opened 2026-09-15
fix: unique insert_id for batched identify transfers (SDKA-117)

ianmeyer:
### Describe what this PR is addressing
Identify batching reuses an already-uploaded `insert_id` when transferring merged `$set` identifies. Ingestion dedups on `insert_id`, so later merged user properties are dropped. Fire-and-forget intercept-file deletes also leave the carrier readable, so later `track()` calls can re-transfer the same event.

Fixes [SDKA-117](https://linear.app/amplitude/issue/SDKA-117/identify-batching-reuses-insert-ids-and-loses-user-properties). Customer-reported (Jubilant FoodWorks / Zendesk 409544); confirmed on analytics-core 1.30.1 and 1.31.0.

### Describe the solution
* Mint a new `insert_id` on every identify transfer (`BaseEvent.withFreshInsertId()`), file and in-memory handlers.
* Await intercept-file removal on the storage I/O dispatcher before returning the transfer event, so the carrier cannot be read again.

No public API change. Identify batching still merges `$set` identifies; transferred events are now unique to ingestion.

### Steps to verify the change
* `./gradlew :analytics-core:test --tests com.amplitude.core.platform.intercept.IdentifyInterceptStorageHandlerTest`
* `./gradlew build ktlintCheck apiCheck -x test` (local CI)

### Useful links and documentation (optional)
* [SDKA-117](https://linear.app/amplitude/issue/SDKA-117/identify-batching-reuses-insert-ids-and-loses-user-properties)
* [Zendesk 409544](https://amplitude.zendesk.com/tickets/409544)

### Checklist
* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-Kotlin/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* [ ] Does your PR have a breaking change?

PR #503, opened 2026-09-15
fix: consume identify batches exactly once (SDKA-117)

ianmeyer:
Follow-up to #502. Identify transfers could overlap while file reads and deletes suspended, allowing an already-emitted intercept batch to absorb later properties and be emitted again with the same `insert_id`.

The first commit reverts #502. Minting a new ID prevented ingestion deduplication but masked duplicate batch emission and increased identify volume.

### Describe the solution
* Serialize identify writes, transfers, identity changes, and `$clearAll` behind one interceptor `Mutex`.
* Treat transfer as a consume operation: later identifies are written only after the prior batch is no longer eligible for merging.
* Preserve the first intercepted event's stable `insert_id` for the merged batch and its upload retries.
* Await file deletion and retain failed deletions in a consumed-file set so cleanup failures cannot make an emitted batch eligible again during the process lifetime.
* Use compare-and-set when scheduling timed transfers.

This fixes runtime batch ownership. It does not claim atomic durability across the intercept and destination stores; the existing crash window before destination persistence remains.

* `./gradlew :analytics-core:test`
* `./gradlew apiCheck`
* `./gradlew build -x test`

Added coverage for sequential batch consumption, stable IDs, overlapping transfers, writes during a suspended transfer, and deletion failure.

* [SDKA-117](https://linear.app/amplitude/issue/SDKA-117/identify-batching-reuses-insert-ids-and-loses-user-properties)
* #502

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes identify batching, locking, and insert_id semantics on a hot analytics path; mistakes could duplicate events or drop merged user properties, though behavior is heavily tested.
> 
> **Overview**
> Fixes overlapping identify transfers that could re-emit the same batch (same `insert_id`) and absorb later properties by **reverting the “fresh insert id on transfer” workaround** and treating transfer as a single consume-and-enqueue step instead.
> 
> **IdentifyInterceptor** now serializes intercepts, transfers, identity changes, and clears behind a **`Mutex`**, schedules batched transfers with **compare-and-set**, and runs **consume + `enqueuePipeline` under `NonCancellable`** so cancellation cannot skip enqueue after a batch is read.
> 
> **File-backed intercept storage** tracks **consumed file paths** so failed deletes cannot make an already-emitted batch eligible again; merge failures **`releaseFile`** instead of deleting; and **`clearIdentifyIntercepts`** marks everything consumed before cleanup. Transfers keep the **first intercepted event’s `insert_id`** (in-memory handler aligned). **`withFreshInsertId`** is removed.
> 
> Tests cover sequential batch consumption, stable IDs, overlapping transfers, writes during transfer, deletion/merge failures, and new concurrency scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 243fdf080a56d48d8fe01072646ff456cff9b3b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crleona:
Bugbot run

analytics-core/src/main/java/com/amplitude/core/platform/intercept/IdentifyInterceptFileStorageHandler.kt
```diff
@@ -17,55 +18,77 @@ public class IdentifyInterceptFileStorageHandler(
...
+                        } else {
+                            val mergedUserProperties = IdentifyInterceptorUtil.mergeIdentifyList(eventsList)
+                            identifyEventUserProperties?.putAll(mergedUserProperties)
+                        }
+                    }
+                processed.onSuccess {
+                    processedFilePaths.add(eventPath)
+                }.onFailure { e ->
+                    logger.warn("Identify Merge error: ${e.message}")
+                    storage.releaseFile(eventPath)
```
crleona: Required: We should not leave this file eligible for a later merge after allowing subsequent events through. If reading user A's batch fails during an identity change, we accept user B's properties and can later merge both files using A's identity. Even for one user, replaying an older batch can overwrite newer values.

For this fix, please preserve the previous discard-on-read/merge-error behavior: log the failure and mark the file consumed so deletion failure cannot reintroduce it. Keep cancellation separate. Please replace the retry-on-merge-failure test with coverage that a failed old batch cannot join a later batch, including across an identity change.

analytics-core/src/main/java/com/amplitude/core/platform/intercept/IdentifyInterceptFileStorageHandler.kt
```diff
@@ -17,55 +18,77 @@ public class IdentifyInterceptFileStorageHandler(
...
+                        val eventsString = storage.getEventsString(eventPath)
+                        if (eventsString.isEmpty()) {
+                            return@runCatchingCancellable
+                        }
+                        val eventsList = JSONArray(eventsString).toEvents()
+                        if (eventsList.isEmpty()) {
+                            return@runCatchingCancellable
+                        }
+
+                        if (event == null) {
```
crleona, line 64: Optional simplification: Can we merge the entire eventsList once, then select its first event as the carrier only if we do not have one yet? The first file does not need separate property-merging logic. Compute the file's merged properties before updating the batch accumulator so a failed merge leaves the accumulator unchanged.

analytics-core/src/main/java/com/amplitude/core/platform/intercept/IdentifyInterceptFileStorageHandler.kt
```diff
@@ -17,55 +18,77 @@ public class IdentifyInterceptFileStorageHandler(
...
             if (e is FileNotFoundException) {
                 e.message?.let { logger.warn("Event storage file not found: $it") }
-                return null
+            } else {
+                throw e
             }
-            throw e
         }
-        val eventsData = storage.readEventsContent()
+        cleanupConsumedFiles()
```
crleona: Optional simplification: We already exclude consumed paths from reads, so can we retry cleanup just once at the end? Filter existing consumed paths, process this batch, mark its paths consumed, then clean up. Ensure the empty-batch path still retries cleanup. The same simplification applies to clearIdentifyIntercepts(), which currently cleans up twice.

crleona (changes requested):
The mutex, stable insert ID, and consumed-file tracking are the right shape and address the reported repeated-batch failure. Keep batching consecutive pending $set identifies; once transferred, that batch must never participate in merging again. The remaining required correction is the read-failure path: retaining failed files while allowing subsequent events through breaks ordering and can merge properties across identities. The other two comments are optional simplifications.

crleona (approved):
LGTM!

PR #504, opened 2026-09-17
feat: add Amplitude.untrackPlayer to stop tracking a Media3 player

ianmeyer:
Apps need a way to end Streaming Analytics on a Media3 player immediately (for example before `release()`), and terminal Stream Stopped events were racy: async `stop()` could overlap a re-track, and plugin teardown could drop delayed events because the enrichment plugin is removed before `teardown()` runs.

### Describe the solution
* Add `Amplitude.untrackPlayer(player)`. The player is already held with a `WeakReference`, so this is optional for leaks and only needed for an immediate terminal stream/ad stop.
* `PlayerBinding.stop()` owns `finishPlayback` on a parentless cleanup scope; `stopAndJoin()` only joins that job, so cancelling `untrackPlayer` or the graph cannot abort the terminal event.
* Factory mutex serializes detach vs getOrCreate; `detachAll()` closes the factory so teardown cannot open a new session afterward.
* Teardown routes delayed events into `storagePipeline` (honoring `optOut`) and flushes uploads, because the plugin is already off the timeline.

* `./gradlew build -x test && ./gradlew test ktlintCheck apiCheck`
* Unit tests covering untrack, detach/re-track ordering, cancelled detach still emitting Stream Stopped, teardown opt-out, and orphan unregister.

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes async player attach/detach, terminal event delivery, and plugin teardown ordering; mistakes could drop or mis-route Stream Stopped events, though coverage targets ordering, cancellation, and teardown routing.
> 
> **Overview**
> Adds **`Amplitude.untrackPlayer(player)`** so apps can end Streaming Analytics on a Media3 player immediately (e.g. before `release()`), emitting a terminal **Stream Stopped** with `stop_reason` **untracked** when a session is open. **`trackPlayer`** and **`untrackPlayer`** are serialized with a global mutex so they apply in call order even though both run asynchronously.
> 
> **Player lifecycle** changes: **`PlayerBindingFactory`** uses a coroutine mutex, **`detach` / `detachAll`**, and a **closed** flag so teardown cannot open new sessions after drain. **`PlayerBinding.stop()`** runs terminal cleanup on a parentless scope and **`stopAndJoin()`** waits for that job so cancelling the caller cannot drop the stop event.
> 
> **Plugin teardown** routes stream **DelayedEvent**s through **`StreamTracker.routeDelayedEvents`** into **`storagePipeline`** (with opt-out and Before-plugin enrichment) because the enrichment plugin is removed from the timeline before **`teardown()`** runs, then flushes uploads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3fb2603896962ff3ea23d6abc1449f9d3f8a0ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
```diff
@@ -30,14 +35,40 @@ public fun Amplitude.trackPlayer(
...
+ * Safe to call more than once.
+ *
+ * ```
+ * amplitude.untrackPlayer(exoPlayer)
+ * exoPlayer.release()
+ * ```
+ */
+@AmplitudePreview
+public fun Amplitude.untrackPlayer(player: Player) {
+    amplitudeScope.launch(amplitudeDispatcher) {
```
polbins: It seems both track/untrack use the same `Dispatchers.IO` so it could have an ordering race. Could we ensure the call order is respected here?

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/player/PlayerBinding.kt
```diff
@@ -92,12 +94,17 @@ internal class PlayerBinding internal constructor(
...
 
-    fun stop() {
-        if (!stopped.compareAndSet(false, true)) return
+    /**
+     * Starts the terminal stop once and returns a job completing when its events are tracked.
+     *
+     * The stop runs on a scope of its own, independent of both the graph and the caller, so
+     * cancelling either cannot leave a stream session without its Stream Stopped.
+     */
+    fun stop(): Job {
```
polbins, line 103: should we wrap this with `runCatching`?

ianmeyer: I'll look. I think real work is pretty much inside the try/catch.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
ianmeyer: Sure. I ignored this previously as not really a race we expect to impact us, but I can make it more robust.

streaming-analytics-android/src/main/kotlin/com/amplitude/android/streaming/internal/player/PlayerBinding.kt
ianmeyer: I don't think there's value in wrapping this further

streaming-analytics-android/src/main/kotlin/com/amplitude/android/TrackPlayer.kt
ianmeyer: addressed

polbins (approved):

PR #509, opened 2026-09-18
feat: add unified Android SDK wrapper

polbins:

Implements SDKA-12, a single Android entry point that installs Analytics, Session Replay, Experiment, and Guides & Surveys with shared host state and nullable blade accessors.

### Describe the solution

- Adds the thin com.amplitude:unified-android module with api dependencies on the four SDKs.
- Adds UnifiedConfigurationBuilder and AmplitudeUnified, which inherits the Analytics Android API.
- Installs blades in Session Replay, Experiment, then Guides & Surveys order.
- Keeps Analytics usable when an individual blade fails during setup.
- Makes Experiment access return null for ambiguous multi-instance registration.
- Applies unified library attribution after the Android context plugin and preserves existing library attribution.
- Pins released compatible artifacts: Analytics 1.30.1, Experiment 1.17.0, Session Replay 0.30.0, and Engagement 3.14.0.
- Adds Kotlin and Java unit coverage for configuration, accessors, failure isolation, attribution, real blade construction, and Java construction.

- gagent :unified:testDebugUnitTest
- gagent ktlintCheck
- gagent apiCheck
- gagent build -x test

All pass locally.

- [SDKA-12](https://linear.app/amplitude/issue/SDKA-12/unified-amplitude-wrapper)
- SDK verification updates are staged separately in the amp-kotlin-sdk-verification-phase-1-5 repository.

### Checklist


gradle/libs.versions.toml
```diff
@@ -14,6 +14,10 @@ curtains = "1.2.5"
 dokka = "2.0.0"
 espressoCore = "3.7.0"
 experimentAndroidClient = "1.6.3"
+# Replace these prototype versions with the first released UniversalPlugin versions before publishing.
+unifiedExperimentAndroidClient = "1.17.0"
+sessionReplayAndroid = "0.30.0"
+engagementAndroid = "3.14.0"
```
polbins, line 20: this is still WIP, we're just assuming that the plugin is going to be released with this version

gradle/libs.versions.toml
```diff
@@ -14,6 +14,10 @@ curtains = "1.2.5"
 dokka = "2.0.0"
 espressoCore = "3.7.0"
 experimentAndroidClient = "1.6.3"
+# Replace these prototype versions with the first released UniversalPlugin versions before publishing.
+unifiedExperimentAndroidClient = "1.17.0"
+sessionReplayAndroid = "0.30.0"
```
ianmeyer: unified and core share the same version. if that version is ever binary-incompatible with SR defined here, then unified breaks at runtime - until SR is re-released and then unified is re-released with a bumped version of SR.

We'll have to make sure we don't ship broken versions of unified.

gradle/libs.versions.toml
ianmeyer: the blades - SR, engagement, experiment – depend on core/android. At some point, we will publish a version of core/android that breaks those integrations.  If unified is published alongside core/android, then eventually we will publish a version of unified that is broken. I think unified should have a separate release and publish workflow, so we can release updated blades before releasing an updated unified.

gradle/libs.versions.toml
```diff
@@ -38,6 +38,8 @@ robolectric = "4.15.1"
 testCore = "1.7.0"
 testExtJunit = "1.3.0"
 testRunner = "1.7.0"
+unifiedExperimentAndroidClient = "1.17.0"
```
ianmeyer, line 42: I think the `analytics-core` / `analytics-android` versions should also be pinned here, or we would be rebuilding them when we release unified.

package-lock.json
```diff
@@ -1,14 +1,16 @@
 {
-  "name": "Amplitude-Kotlin",
+  "name": "amp-kotlin-sdka12-wrapper",
```
ianmeyer: what is the effect of changing this name to `amp-kotlin-sdka12-wrapper`?

unified/build.gradle.kts
```diff
@@ -0,0 +1,140 @@
...
+    testRuntimeOnly(libs.junit.jupiter.engine)
+    testRuntimeOnly(libs.junit.vintage.engine)
+    testRuntimeOnly(libs.junit.platform.launcher)
+    testImplementation(libs.junit4)
+    testImplementation(libs.robolectric)
+    testImplementation(libs.test.core)
+}
+
+// Unified releases on its own cadence (release-unified.yml), so core releases must not publish it.
+val isUnifiedRelease = providers.gradleProperty("amplitude.unified.release").isPresent
```
ianmeyer: `.isPresent` means that the value is not important, any value will make `isUnifiedRelease` true.

unified/build.gradle.kts
```diff
@@ -0,0 +1,140 @@
...
+        AndroidSingleVariantLibrary(
+            variant = "release",
+            sourcesJar = true,
+            publishJavadocJar = true,
+        ),
+    )
+}
+
+dependencies {
+    api(project(":android"))
```
ianmeyer: same comment as earlier. depending on the released binary, `api(libs.unified.analytics.android)` seems safer here than the current code.

package-lock.json
polbins: It's unintended and it's just my local worktree name, fixing.

ianmeyer:
PR title should be changed to `feat(unified):` right? 😄

polbins:
> PR title should be changed to `feat(unified):` right? 😄

Ahh, you're right! Let me update it.

ianmeyer (approved):

